PR #5, opened 2017-02-24
Add extensionpoint to sharedialogview

felixheidecke:
Write method overriding extension

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -45,6 +45,8 @@ if (in_array('contact', $conditions)) {
 \OCP\Util::connectHook('OC_Filesystem', 'preSetup', '\OCA\Guests\Hooks', 'preSetup');
 
 // --- register js for user management------------------------------------------
+\OCP\Util::addScript('guests', 'app');
+
 $user = \OC::$server->getUserSession()->getUser();
 if ($user) {
     // if the whitelist is used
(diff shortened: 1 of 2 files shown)


PVince81:
Works for me, I put this line:
```js
OC.Plugins.register('OCA.Share.foo');
```

at the end of ShareDialogView.initialize().

A breakpoint inside https://github.com/owncloud/guests/pull/5/files#diff-41d794d24ff042b1f9ac211fc3f9f951R4 properly stops when I type something in the autocomplete dialog.

The problem you probably have is that `this` is not defined the same.
You need to use `obj` instead.
Or define an outside variable `var self = obj;` before line 3 here https://github.com/owncloud/guests/pull/5/files#diff-41d794d24ff042b1f9ac211fc3f9f951R3.

PVince81:
ah, or replace `var view = this;` with `var view = obj;` here https://github.com/owncloud/guests/pull/5/files#diff-41d794d24ff042b1f9ac211fc3f9f951R4

And `this.$el.` with `view.$el.`

PVince81:
Or an even more advanced technique, add this at the end of `attach`:

```js
_.bind(obj.autocompleteHandler, obj);
```

This will make sure that `this` is actually `obj` when used inside your `autocompleteHandler`.

felixheidecke:
@PVince81 **thx!** I didn't realize that `this` wouldn't refer to the current object :sweat_smile:

> Works for me, I put this line:
> OC.Plugins.register('OCA.Share.foo');
> at the end of ShareDialogView.initialize().

I think you forgot to pass `this` in the register method, otherwise, there will be nothing to extend :clap:

PR #44, opened 2017-03-30
Starting integration tests for guests app.

SergioBertolinSG:


SergioBertolinSG:
Next thing to add is the deletion of all guest users created during the tests. This includes a function to delete an specific guest user. This could require a function to get the id of the guest  if it cannot be deleted by any of the parameters used to create it. Also a function to get all the guests in the system.

SergioBertolinSG:
Requires this PR to merged in core to work. 
https://github.com/owncloud/core/pull/27618

Makefile
```diff
@@ -0,0 +1,55 @@
...
+clean_rules=
+js_rules=
+help_rules=help-base
+
+tools_path=$(shell pwd)/tools
+
+.PHONY: all
+all: help-hint dist
+
+include rules/deps.mk
```
PVince81: you are missing all these other files.

If all you wanted to do is run the tests from travis then I suggest to simply add the "run.sh" command directly in travis.

SergioBertolinSG: Yes, better and add the makefile and other stuff when building and unit tests are made.

PR #49, opened 2017-04-05
Retreive email trough interface instead of directly from db

IljaN:
The guest app mailer used to retreive the users email address directly from oc_preferences. With the changes to account table this doesn`t work anymore. 

This pr modifies the code to use IUserManager instead of accessing the db directly.


PVince81:
:+1: that's the correct way

Issue #60, opened 2017-04-12
Can't share to guest as non-admin


PVince81:
### Steps

1. Login as non-admin
1. Type in a name "test" in the share dialog
1. Type in email address
1. Click Save

```
{"reqId":"KwcNsXkdn00iEWDbQNte","level":0,"time":"2017-04-12T13:12:47+00:00","remoteAddr":"127.0.0.1","user":"user1","app":"no app in context","method":"PUT","url":"\/owncloud\/index.php\/apps\/guests\/users","message":"Logged in user must be an admin"}
```

The endpoint is likely missing the `@NoAdminRequired` annotation

@IljaN @SergioBertolinSG

IljaN:
Tested

PR #61, opened 2017-04-12
Can`t share to guest as non admin #60

IljaN:


IljaN:
Manually tested -> works

Issue #55, opened 2017-04-10
Wrong default permissions


PVince81:
### Steps

1. Share a folder "test" with guest: type in a name "someone" to create a guest share
1. In the popup type email address and save
1. Receive email and set password
1. Login as "someone"
1. Enter folder "test": no permission to write, but "Rename" is visible
1. Log out
1. Login as the sharer
1. Check permissions: checkbox says "can edit"
1. Uncheck then check again the permission "can edit"
1. Login as "someone"
1. Enter folder "test": permissions are correct now

### Expected result
Permissions are correct in the first go.

### Actual result
The UI says the perms are "can edit" but actually it's a different perm.

Setting to critical because this is a bad first experience for a user even though there is a workaround. It makes it feel like the app is broken.

@IljaN please fix

cc @SergioBertolinSG

PVince81:
Still happening for me on master (pulled both core master and guests).

After the first share I see the permissions being "3" in the database, this looks wrong:
```
+----+------------+---------------+-----------+---------------+--------+-----------+-------------+-------------+-------------+-------------+-------------+------------+----------+------------+-------+-----------+------------+
| id | share_type | share_with    | uid_owner | uid_initiator | parent | item_type | item_source | item_target | file_source | file_target | permissions | stime      | accepted | expiration | token | mail_send | share_name |
+----+------------+---------------+-----------+---------------+--------+-----------+-------------+-------------+-------------+-------------+-------------+------------+----------+------------+-------+-----------+------------+
|  1 |          0 | vincent_petry | admin     | admin         |   NULL | folder    | 14          | NULL        |          14 | /test       |           3 | 1492023567 |        0 | NULL       | NULL  |         0 | NULL       |
+----+------------+---------------+-----------+---------------+--------+-----------+-------------+-------------+-------------+-------------+-------------+------------+----------+------------+-------+-----------+------------+
```

If I uncheck the checkbox the permissions in DB become 1.

Then when I tick it again it becomes 15, which is correct.

So the bug is that for some reason the initial permissions is 3.

IljaN:
So just to clarify again these are my initial permissions without touching anything:
![perms](https://cloud.githubusercontent.com/assets/246983/24996298/9cf018e2-2032-11e7-8290-b4ab3d179265.png)

On your side rename is missing despite "Can Edit" beeing checked?

PVince81:
No, the bug is that I can't upload into that folder even though "can edit" is checked.

[...]

PR #66, opened 2017-04-13
Inconsistent default permission 

IljaN:
#55

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -115,7 +115,7 @@ $(document).ready(function () {
 				var attributes = {
 					shareType: 0,
 					shareWith: this.guest.username,
-					permissions: 3,
+					permissions: 15,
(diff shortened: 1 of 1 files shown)


IljaN:
Tested

js/app.js
```diff
@@ -115,7 +115,7 @@ $(document).ready(function () {
 				var attributes = {
 					shareType: 0,
 					shareWith: this.guest.username,
-					permissions: 3,
+					permissions: 15,
```
PVince81: use the constants would be better: OC.PERMISSION_READ | OC.PERMISSION_UPDATE | ... for better readability

IljaN: Oh Okay, didn`t know that they are also available in the frontend....

PR #78, opened 2017-04-26
Add build process for FE code

felixheidecke:
Frontend code needs to be build using **npm** and **webpack**.
[See the Wiki](https://github.com/owncloud/guests/wiki/Frontend-development#building) on how to do that.


PVince81:
Now add a makefile with `make dist` that copies the relevant stuff to "build" like the market app does.
And `make debug` that runs `npm run dev`

PR #83, opened 2017-05-17
Move to authentication section instead of own section

tomneedham:
I think this makes sense in the authentication section. In core we are renaming 'Authentication' to 'User Authentication'. Better than registering its whole own section.


tomneedham:
Core renaming: https://github.com/owncloud/core/pull/27918

PR #85, opened 2017-05-18
Use user id instead of user object in mail status

PVince81:
Prevents warnings in log

- [x] TEST: creating guest user still works (email, set password)
- [x] TEST: share to existing email does not show warning any more (requires and observed in https://github.com/owncloud/core/pull/27906)

@IljaN please review


IljaN:
@PVince81 thxx

PR #98, opened 2017-06-15
Mail already used for other guest user

davitol:
Add integration test mail already used for other guest user

### DIFF
--- a/tests/integration/guests_features/Guests.feature
+++ b/tests/integration/guests_features/Guests.feature
@@ -56,8 +56,8 @@ Scenario: A created guest user can log in
 
 Scenario: Trying to create a guest user that already exists
 	Given As an "admin"
-	When user "admin" creates guest user "guest" with email "guest@example.com"
+	And user "admin" creates guest user "guest" with email "guest@example.com"
 	And the HTTP status code should be "201"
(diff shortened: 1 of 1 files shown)


tests/integration/guests_features/Guests.feature
```diff
@@ -53,3 +53,11 @@ Scenario: Check that skeleton is properly set
 	Then the HTTP status code should be "200"
 	And user "guest_example_com" should see following elements
 		| /textfile1.txt |
+
+Scenario: Trying to create a guest user that already exists
+	Given As an "admin"
+	When user "admin" creates guest user "guest" with email "guest@example.com"
+	And the HTTP status code should be "201"
+	And check that user "guest" is a guest
+	And user "admin" creates guest user "guest" with email "guest@example.com"
```
SergioBertolinSG: This step is the 'When'.

SergioBertolinSG:
The title here doesn't correspond to the test, as it has the same user name also. Not just the same email address.

davitol:
> The title here doesn't correspond to the test, as it has the same user name also. Not just the same email address.

The username is created through the mail

screen shot 2017-06-15 at 15 00 31

PR #99, opened 2017-06-19
guest user uploads and reshares

davitol:
Integration test for upload and reshare


davitol:
Close in favour of https://github.com/owncloud/guests/pull/103

Issue #125, opened 2017-07-21
Cannot save the changes of the whitelist features for guest pressing enter key


davitol:
As in other browsers, it should be possible to save the changes of the whitelist features for guest pressing the enter key. For IE11 the way to save the changes is clicking away from the textbox.

@pmaier1 @cortho FYI

pmaier1:
Yes, good point!

cortho:
PR: https://github.com/owncloud/guests/pull/127

cortho:
fixed with #127

davitol:
Tested and works

Issue #112, opened 2017-07-18
It is possible to erase core app from guests whitelist


davitol:
![screen shot 2017-07-18 at 17 03 28](https://user-images.githubusercontent.com/10596560/28324398-89747340-6bdb-11e7-904e-6380bcefcb65.png)

Editing this line it is possible to delete core app that will make nonsense to have guest users only for get this screen, and cannot deploy no further action.

![screen shot 2017-07-18 at 16 46 44](https://user-images.githubusercontent.com/10596560/28324543-f927a5cc-6bdb-11e7-8610-3ca175726432.png)

@cortho

davitol:
Also if we delete the first comma of the list, or if we add an app name at the beggining, the guests account are no longer available

cortho:
Most probably we should move the comma and core to the backend and have the whitelist only display apps, which actually can be removed.

We'll need this anyway, when we want to implement https://github.com/owncloud/guests/issues/106

pmaier1:
Valid points!

cortho:
PR: https://github.com/owncloud/guests/pull/127

cortho:
fixed with #127

PR #127, opened 2017-07-27
Whitelist: move "," and core to backend, have enter key working

cortho:
* comma and core are moved to the backend and not visible in the settings any more https://github.com/owncloud/guests/issues/125
* accept enter key to save group and whitelist https://github.com/owncloud/guests/issues/112
* removed getWhitelist from settingscontroller and routes as it was not in use

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 guests 
 =======
 
-Create a guest user by typing his name in to the sharing dialog. The guest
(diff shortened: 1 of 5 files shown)


cortho:
I found out that files is essential for a _core whitelist_. Did i miss another app which should/can never be disabled?

pmaier1:
@PVince81 ^

felixboehm:
👍

Issue #24, opened 2017-03-10
Add build script


PVince81:
Like a Makefile where `make dist` builds the app into "build/guests/" and strips all unwanted files.

See https://github.com/owncloud/customgroups/blob/master/Makefile for an example, it has sub-rules-files.
I hope to make this reusable for apps eventually.

PVince81:
@felixheidecke is this app still using LESS/SASS ? We can't rely on people using specific IDEs to build this app especially if we want to automate this at some point.

felixheidecke:
Nope … no pre-compilers at this point.

PVince81:
maybe reuse the makefile from the market app

PVince81:
Can we have a makefile here ? Checking out the app alone doesn't seem to be enough to make it work and I always forget those `npm run build` commands...

felixheidecke:
Well … I'll read into makefiles then :-)

pmaier1:
@cortho will take care, right?

cortho:
PR: #137

PR #137, opened 2017-08-21
add Makefile, adjust README/fix path in integration tests

cortho:
Here is a Makefile, as requested in #24 

I tried to get the appstore/signing procedure into the Makefile as well, but since this is my first for ownCloud, please be considerate with me.

Does it comply with the requirements?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,5 @@
 app_name=$(notdir $(CURDIR))
-source_build_directory=$(CURDIR)/build/artifacts/source
-source_package_name=$(source_build_directory)/$(app_name)
-appstore_build_directory=$(CURDIR)/build/artifacts/appstore
+appstore_build_directory=$(CURDIR)/build/
 appstore_package_name=$(appstore_build_directory)/$(app_name)
 tests_integration_directory=$(CURDIR)/tests/integration
 
@@ -33,7 +31,7 @@ dev: npm
 # Node modules
 .PHONY: npm
 npm:
-	npm install && npm update
+	npm install
 	
 # Remove the appstore build and generated guests bundle
 .PHONY: clean
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,83 @@
+app_name=$(notdir $(CURDIR))
+source_build_directory=$(CURDIR)/build/artifacts/source
+source_package_name=$(source_build_directory)/$(app_name)
+appstore_build_directory=$(CURDIR)/build/artifacts/appstore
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,83 @@
...
+
+# Fetch JS dependencies and compile the JS
+.PHONY: dev
+dev: npm
+	npm run dev
+
+# Node modules
+.PHONY: npm
+npm:
+	npm install && npm update
```
PVince81: Is `npm update` really needed ?

Regarding npm deps, make sure they are locked to specific versions to avoid having strange issues due to unplanned dependency updates.

Makefile
```diff
@@ -0,0 +1,83 @@
+app_name=$(notdir $(CURDIR))
+source_build_directory=$(CURDIR)/build/artifacts/source
+source_package_name=$(source_build_directory)/$(app_name)
+appstore_build_directory=$(CURDIR)/build/artifacts/appstore
```
PVince81: Make that `$(CURDIR)/build`, that should be enough. We have this in other apps too

Makefile
```diff
@@ -0,0 +1,83 @@
+app_name=$(notdir $(CURDIR))
+source_build_directory=$(CURDIR)/build/artifacts/source
```
PVince81: I don't think we need a source build. People should just grab the github version from a tag and run make.

Please remove.

Makefile
cortho: No, I guess, this is not needed. Copied it from contacts app's package.json's prebuild. So I'll remove that?

Makefile
cortho: ack

Makefile
cortho: ok.

cortho:
@PVince81 removed recommended items, for npm there already dependencies in package.json.

Something more, I should consider?

PVince81:
I think it's fine now. Most important is that `make dist` builds the tarball in "build" directory for easy pickup.

I haven't tested signing as I don't have the key, but as you took the code from contacts it should work.

DeepDiver1975:
> I haven't tested signing as I don't have the key, but as you took the code from contacts it should work.

let me test this ....

cortho:
@DeepDiver1975 Thank you. I assumed that _make_ precedes a _make appstore_ added build target to appstore

davitol:
Tested and works. 👍

Issue #84, opened 2017-05-18
Email notification resent when sharing with existing guest by email


PVince81:
The code sends another notification every time you share with a guest using an existing email.

The mail template is different though: it's the generic "hey, a user shared XYZ" with you.
Normally this is the responsibility of the code share code to send these emails **if enabled by the admin**.

This looks like a bug. I think the mail handler in the guest app needs to check whether the guest user was just created and in this case send the mail. Else don't send anything.

@IljaN

cortho:
@PVince81 @pmaier1 
This second email is provided intentionally, see https://github.com/owncloud/guests/blob/master/lib/hooks.php#L196

It is very easy to just remove these lines. Your opinions?

cortho:
Just tested this out. Emails are sent to guests, when admin checks _Allow users to send mail notification for shared files to other users_ in Sharing settings, so I'll remove this functionality from the guests app.

pmaier1:
> Just tested this out. Emails are sent to guests, when admin checks Allow users to send mail notification for shared files to other users in Sharing settings, so I'll remove this functionality from the guests app.

Fine :ok_hand: Share notification mails will still be sent when the settings is enabled, right?

cortho:
> Share notification mails will still be sent when the setting is enabled, right?

yes, and only then

cortho:
PR here: #139

cortho:
fixed with #139

PR #139, opened 2017-08-22
do not send share information after guest had registered

cortho:
Removed email functionality in post_shared hook when guest user is already registered (as discussed in #84).

What was tested?
Do not send w/o admin settings _Allow users to send mail notification ..._:
 * Share with a new guest user -> invite email is sent
 * Share another file with this guest user -> no email is sent

Send with admin settings
 * As an Admin in the Sharing Section check _Allow users to send mail notification for shared files to other users_
 * Share with an existing guest user
 * Check _notify by email_ in the ShareTab -> email is sent

### DIFF
--- a/lib/hooks.php
+++ b/lib/hooks.php
@@ -190,20 +190,6 @@ class Hooks {
 					$itemSource,
 					$registerToken
 				);
-			} else {
-				// always notify guests of new files
-				$guest = $this->userManager->get($shareWith);
-
-				if (!$guest) {
-					throw new DoesNotExistException("$shareWith does not exist");
-				}
-
-				$this->mail->sendShareNotification(
-					$this->userSession->getUser(),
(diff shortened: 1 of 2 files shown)


lib/hooks.php
```diff
@@ -190,20 +190,6 @@ public function handlePostShare(
 					$itemSource,
 					$registerToken
 				);
-			} else {
-				// always notify guests of new files
```
PVince81: @butonic seems this is something you decided to add in the first place ?

butonic: yes, was a requirement at the time to always notify guests of changes. If we change it, it needs to be mentioned in the changelog.

PVince81: I think at this stage, guests could setup the activity app to receive notifications like regular users

PR #146, opened 2017-09-06
Release Changes for master

felixboehm:
- better Makefile
- added .gitignore

Open tasks for using default Makefile
- [ ] add webpack build to default Makefile 
- [ ] add testing to default Makefile
- [ ] update README.md on how to use Makefile
- [ ] npm dependencies cleanup?
  - vue -> production dependency
  - rest -> dev dependency

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -22,7 +22,7 @@ Furthermore guests are fully auditable with the [ownCloud Auditing application](
 	<documentation>
     		<user>https://doc.owncloud.com/server/10.0/user_manual/files/webgui/sharing.html?highlight=guest#sharing-files-with-guest-users</user>
     	</documentation>
-	<licence>AGPL</licence>
+	<licence>GPLv2</licence>
 	<author>ownCloud GmbH</author>
 	<dependencies>
 		<owncloud min-version="10.0" max-version="10.0" />
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 /.idea
 /node_modules
+/build/
 *.bundle.js
 npm-debug.log
+package-lock.json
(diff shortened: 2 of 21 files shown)


appinfo/info.xml
```diff
@@ -22,7 +22,7 @@ Furthermore guests are fully auditable with the [ownCloud Auditing application](
 	<documentation>
     		<user>https://doc.owncloud.com/server/10.0/user_manual/files/webgui/sharing.html?highlight=guest#sharing-files-with-guest-users</user>
     	</documentation>
-	<licence>AGPL</licence>
+	<licence>GPLv2.0</licence>
```
DeepDiver1975: needs to be added to marketplace first

PVince81:
Better Makefile but not as good as the other ones (this one seems to be different). Running `make` twice will rerun npm install every time due to missing a "package.json" related rule.

Doesn't matter, as long as `make dist` works it's fine for now.

See comment from @DeepDiver1975 before merging.

appinfo/info.xml
felixboehm: is wip

felixboehm:
Please keep release branch, until we decide how to persist release version.

appinfo/info.xml
IljaN: @felixboehm can you please use GPLv2 as license key to be in line with other license types? GPLv2 is also the official abbreviation accorind to https://www.gnu.org/licenses/gpl-2.0.html

IljaN:
If we want to release this under GPLv2 we should change the license headers in all of the source files

DeepDiver1975:
> If we want to release this under GPLv2 we should change the license headers in all of the source files

@IljaN done

PR #149, opened 2017-09-13
remove modal and vue.js, create guests only by email

### LINKED ISSUE #115: Guest users cannot be distinguished from regular users in the sharingTabView
Guest users cannot be distinguished from regular users in the sharingTabView
screen shot 2017-07-18 at 17 19 59

In the screenshot dtoledo is already a guest user, but no info in  in the sharingTabView let us know that he is a guest.
 
@pmaier1 FYI @cortho

### LINKED ISSUE #147: Remove guest creation modal
Currently the modal that opens after clicking 'Add...' in the sharing dialog has only two purposes: Setting a user name and specifying an email address when just a name has been entered. We should give up this part of the feature in favor of making guest creation require one step less.

- Remove the ability to enter a user name for guest creation in favor of just entering mail addresses
- When address is not existing -> create guest
- When address is existing -> directly share internally (needs https://github.com/owncloud/guests/issues/115)

cortho:
**Related issues:**
#147, ~~#115~~

I have been testing locally at the frontend and with the integration tests.

**What is done in this PR:**
- moved dynamically created guests.bundle.js to guestshare.js
- removed modal window (fixes #147)
- ~~extended \Files_Sharing\Controller\ShareesController to be able to distinguish between guests and oc users (#115)~~
- use constant for default group name
- ~~show (guest) in autocomplete, when user is a guest (this value will be taken from appconfig's group name)~~
- ~~do not offer guest user creation in autocomplete when an email is already in use~~
- removed vue.js along with composer & co
- removed build information from Makefile

Edit: current status


js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+
+var Guests = {
+	model: null,
+	email: null,
+	
+	addGuest: function (model, email) {
+		this.model = model;
+		this.email = email;
+		
+		let self = this;
```
PVince81: does `let` work with IE11 without transpiler ?

PVince81: and other browsers like latest Safari, etc... I'd prefer to be safe and use `var` instead

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+				email: this.email
+			}
+		}
+
+		$.ajax(xhrObject).done(function (xhr) {
+			self._addGuestShare();
+		}).fail(function (xhr) {
+			let response = JSON.parse(xhr.responseText);
+			let error = response.errorMessages;
+			OCdialogs.alert(
```
PVince81: optional: OC.dialogs.alert (OC.dialogs is an alias of the internal OCdialogs)

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+			path: this.model.fileInfoModel.getFullPath()
+		}
+
+		return $.ajax({
+			type: 'POST',
+			url: OC.linkToOCS('apps/files_sharing/api/v1', 2) + 'shares?format=json',
+			data: attributes,
+			dataType: 'json'
+		}).done(function () {
+			if (self.model)
```
PVince81: please add brackets even for a single line

see https://doc.owncloud.org/server/10.0/developer_manual/general/codingguidelines.html#control-structures

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+				t('core', 'Error') // title
+			);
+		});
+	},
+}
+
+OC.Plugins.register('OC.Share.ShareDialogView', {
+	attach: function (obj) {
+
+		// Override ShareDigalogView
```
PVince81: Digalog

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+				{
+					format: 'json',
+					search: search.term.trim(),
+					perPage: 200,
+					itemType: view.model.get('itemType')
+				},
+				function (result) {
+					$loading.addClass('hidden');
+					$loading.removeClass('inlineblock');
+					if (result.ocs.meta.statuscode == 100) {
```
PVince81: optional: use "api/v2" for slightly more status code sanity and receive 200 here instead of 100

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+var Guests = {
```
PVince81: class is called `Guests` but the file is called `sharedialogview.js` ?

PVince81:
I also need to think about it a bit... my gut feeling tells me that doing "do not offer guest user creation in autocomplete when an email is already in use" is not trivial and might need some backend work instead.

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+		});
+	},
+}
+
+OC.Plugins.register('OC.Share.ShareDialogView', {
+	attach: function (obj) {
+
+		// Override ShareDigalogView
+
+		obj.autocompleteHandler = function (search, response) {
```
PVince81: ok now I realize that this is the old existing code that you just moved here... argh... I wish I saw this earlier

js/sharedialogview.js
```diff
@@ -0,0 +1,269 @@
...
+
+						var usersLength;
+						var groupsLength;
+						var remotesLength;
+
+						var i, j;
+
+						// Add potential guests to the suggestions
+						if (searchTerm.search("@") !== -1) {
+							groupName = groupName = "(gast)";
```
PVince81: `(gast)` would need to be translated: `t('guests', '(guest)')`

PVince81: and syntax problem, there are two equals. Change this row to:
```js
groupName = groupName += ' ' + t('guests', '(guest)');
```

js/sharedialogview.js
PVince81: actually what I expected from the original dev was to **extend** the function, not copy paste it.
To extend:
```js
var oldHandler = obj.autocompleteHandler;
obj.autocompleteHandler = function(search, response) {
    return oldHandler.call(obj, search, function() {
        // add additional entries here (challenge might be to get it sorted properly)
        // ...
        response();
    });
}
```

PVince81: We should probably add more hooks in core for the autocomplete handler to allow injecting results inside the array from separate/different requests so they can be sorted and also the "no users" message would work too.

Trying to think of a way without additional hooks...

PVince81: Raised here https://github.com/owncloud/guests/issues/150

PVince81:
Sorry if my comments might sounds confusing. I commented as soon as I saw the code.
I'm starting to get a fuller picture of what you are trying to achieve.

From what I understand:
- ShareesController was overridden to be able to append "($guestGroupName)" for any guest user in the result list
- client-side JS is just a copy of the old vue.js plugin registration, the only change is the endpoint that goes to the guest app's sharee endpoint instead of the core one, to reach ShareesController
- the client-side autocompleteHandler override is only here to inject the extra "create new guest" entry in case an email address was entered.

I'd like to find a way that removes the need to override/extend the ShareesController.
On the other hand, one might say, if we already copy-pasted so much code for the JS override, why not do the same for the server ? :-/

PVince81:
Ideal would to be to have the distinction possible in core, we have been discussing this here: https://github.com/owncloud/core/issues/28979

PVince81:
@cortho I had a chat with @pmaier1 and we agreed that the suffix "($guestGroupName)" is not needed for this version. We'll look into a way for distinguishing users in the share autocomplete directly in core: https://github.com/owncloud/core/issues/28979

Can you remove `ShareesController` and the custom route ?
Removing the dialog and switching to email should be enough for now.

cortho:
@PVince81 ok. I'll look over this and give feedback/changes ASAP

cortho:
@PVince81 I removed SharesController and we are now extending the autocomplete handler, plus the cosmetic/syntax fixes from above. The PR is much smaller now.

js/guestshare.js
```diff
@@ -0,0 +1,136 @@
...
+		var xhrObject = {
+			type: 'PUT',
+			url: OC.generateUrl('/apps/guests/users'),
+			dataType: 'text',
+			data: {
+				displayName: this.email,
+				username: this.email,
+				email: this.email
+			}
+		}
```
PVince81: I suggest running JSLint/JSHint on your JS code. It would tell you to add a semicolon here.

js/guestshare.js
```diff
@@ -0,0 +1,136 @@
...
+					result.push({
+						label: t('core', 'Add {unknown} (guest)', {unknown: searchTerm}),
+						value: {
+							shareType: 4,
+							shareWith: searchTerm
+						}
+					});
+				}
+
+				response(result);
```
PVince81: so it did work, that is really awesome !

Did you notice any sort order issue while testing ? From what I see the original handler sorts before calling the callback: https://github.com/owncloud/core/blob/master/core/js/sharedialogview.js#L248. So you might need to sort again here. Since the list is usually not too long it's ok to do it again.

js/guestshare.js
```diff
@@ -0,0 +1,136 @@
...
+OC.Plugins.register('OC.Share.ShareDialogView', {
+	attach: function (obj) {
+
+		// Override ShareDialogView
+		var oldHandler = obj.autocompleteHandler;
+		obj.autocompleteHandler = function(search, response) {
+
+		    return oldHandler.call(obj, search, function(result) {
+
+				var searchTerm = search.term.trim();
```
PVince81: please check if `result` is defined, if not, call `response()`. See https://github.com/owncloud/core/blob/master/core/js/sharedialogview.js#L269 for an example when it can happen.

cortho:
@PVince81 Thank you for your good review. It looks like we're done then.

PVince81:
hmmm weird

```
± % git checkout -b cortho-new_sharetab_behavior master
git pull https://github.com/cortho/guests.git new_sharetab_behavior

Switched to a new branch 'cortho-new_sharetab_behavior'
From https://github.com/cortho/guests
 * branch            new_sharetab_behavior -> FETCH_HEAD
fatal: refusing to merge unrelated histories
```

PVince81:
wrong repo 🤦‍♂️

PVince81:
@cortho thanks a lot, I'll test it a bit

PVince81:
- [x] TEST: share with email, create guest, account works
- [x] TEST: share with existing guest by entering email address

- [ ] BUG: if I type in the email address of an existing user, the user appears both as "Add guest" and as existing user

@cortho to fix this, I suggest that in your autocompleteHandler, iterate over the results and check if there is any result with the same email address. If yes, then don't add "Add guests".

- [ ] BUG: disable federated sharing (settings/Sharing, untick both under "Federated Cloud Sharing) then share with a new email address like "mail@example.com" (make sure there is no other result). The dropdown does not appear.

Not sure how to fix this. The former autocomplete handler had no results in the set and decided to not display the dropdown, so adding an entry afterwards doesn't properly re-set it.

@cortho would be good if you could fix at least the first issue above. The second one might need dirty hacks, not sure if it will work.

js/guestshare.js
```diff
@@ -0,0 +1,142 @@
...
+					result.push({
+						label: t('core', 'Add {unknown} (guest)', {unknown: searchTerm}),
+						value: {
+							shareType: 4,
+							shareWith: searchTerm
+						}
+					});
+				}
+
+				if (result !== undefined) {
```
PVince81: please move this check at the begining of the function. Because if `result` is undefined, the `result.push` above will fail hard

PVince81:
@cortho another trick to try: instead of adding "Add guest" result inside the overridden handler, add it to the results returned by the server **before** calling oldAutocompleteHandler. Might help solving the second issue and will also resolve the double sorting issue.

PVince81:
I need to finish this, merging for now. Almost there.

PVince81:
the travis tests fail for some strange reason, need to investigate separately

PVince81:
Follow up here: https://github.com/owncloud/guests/pull/151

felixboehm:
This PR removed vue.js, which we agreed on very clear to use as FE framework for this app.
Also the functionality is completely broken, app whitelist is not working.

@cortho

felixboehm:
Question here about vue.js https://github.com/owncloud/guests/issues/147#issuecomment-328856023

IMHO: Even without modal, the vue.js is still needed. So do not remove it.

PR #151, opened 2017-09-14
Fix guest autocomplete addition

PVince81:
Only add guest entry if there is no matching user share in the list.

Detect if no results exist and adjust the UI accordingly when adding
guest entry.

Follow up of https://github.com/owncloud/guests/pull/149

### DIFF
--- a/js/guestshare.js
+++ b/js/guestshare.js
@@ -91,23 +91,37 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 
 				// Add potential guests to the suggestions
 				if (searchTerm.search("@") !== -1) {
-					result.push({
-						label: t('core', 'Add {unknown} (guest)', {unknown: searchTerm}),
-						value: {
-							shareType: 4,
-							shareWith: searchTerm
-						}
-					});
-				}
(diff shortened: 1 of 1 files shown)


PVince81:
See test cases here https://github.com/owncloud/guests/pull/149#issuecomment-329529880.
The two mentioned bugs are fixed here.

Integration tests do not run properly for me locally (`cURL error 7: Failed to connect to 127.0.0.1 port 8025: Connection refused (GuzzleHttp\Exception\ConnectException)`), seems they are not properly portable ? @SergioBertolinSG 

Anyway, let's see what Travis says as it failed on the former PR.

PVince81:
Seems I need to manually start the mailhog docker, I don't like this, should be part of the script. Raised here: https://github.com/owncloud/guests/issues/152

PVince81:
Tested and approved by @davitol

PVince81:
weird travis failure: `Class GuestsContext does not have a constructor, so you cannot pass any constructor arguments`

Will look into this separetely

PVince81:
(the tests ran fine for me locally)

PVince81:
raised travis issue here: https://github.com/owncloud/guests/issues/153

Issue #163, opened 2017-10-04
app whitelist not hiding navigation and tab entries


felixboehm:
- js/navigation.js is not loaded at all
- route apps/guest/apps is not existing
- tabs are not handled at all

Please reproduce by removing the activity app from whitelist.

cortho:
looks like this was lost in https://github.com/owncloud/guests/commit/38278055631b51a8e444369ccc864cc38cf58796#diff-0e033fd8797499a87a6bf2ffccef5c3c and if tabs never were handled ... starting to get it back/fix

cortho:
I fixed the first two issues on my local environment and would be ready to create a PR for this (= displayed apps in the app menu). While fixing and looking deeper into the apps, some questions about non-whitelisted apps arose:

1. At least the _comments_ app does not call URLs like /index.php/apps/gallery but /remote.php/dav/comments. This must be considered at https://github.com/owncloud/guests/blob/master/lib/appwhitelist.php#L80 as well. Are there any other apps, which use a different pattern?

2. There are different approaches how apps connect to the frontend, I'll take _activity_, _favorites_, _files_texteditor_ and _gallery_ as examples;
- _activity_ provides a tab
- _favorites_ are accessible via a menu button.
- _files_texteditor_ binds to OCA.Files.fileActions
- _gallery_ creates a button above the file list

I'm not sure, if there is functionality to unregister listeners etc. for all of the above variants, so what do you think is the best approach? Maybe @PVince81 or somebody else is firm with all of the internals?

3. With these insights, I guess we could cover all community and enterprise apps but we'll never be able to handle all third party apps on the frontend?

pmaier1:
Related: https://github.com/owncloud/guests/issues/136

cortho:
According to the call with @felixboehm I created PR https://github.com/owncloud/guests/pull/165
I'd prefer not to close this issue too soon, as long as this topic has not been discussed.
We might want to merge with #136, hovever and close one of the issues.

pmaier1:
This can be closed due to https://github.com/owncloud/guests/pull/165, right?

cortho:
I'd rather not close this issue too soon, because we still have open questions (https://github.com/owncloud/guests/issues/163#issuecomment-334447344), but closing #136 which is addressing the same basic issue.

pmaier1:
Moving to the next version as we need a release now.

PVince81:
moving to p3, it's in "planned" still

davitol:
[...]

PR #165, opened 2017-10-09
Apply whitelist to frontend and comments URLs

cortho:
Inspired by https://github.com/owncloud/guests/issues/163 we bring back basic frontend app whitelistening:
* remove non-whitelisted apps from navigation menu
* cosmetic removal of items in files view (side nav, buttons, tabs)
* block /dav/comments if not whitelisted

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -34,7 +34,8 @@ $eventDispatcher->addListener(
 $config = \OC::$server->getConfig();
 $groupName = $config->getAppValue('guests', 'group', \OCA\Guests\GroupBackend::DEFAULT_NAME);
 
-\OC::$server->getGroupManager()->addBackend(new \OCA\Guests\GroupBackend($groupName));
+$groupBackend = new \OCA\Guests\GroupBackend($groupName);
+\OC::$server->getGroupManager()->addBackend($groupBackend);
 \OCP\Util::connectHook('OCP\Share', 'post_shared', '\OCA\Guests\Hooks', 'postShareHook');
 
 $user = \OC::$server->getUserSession()->getUser();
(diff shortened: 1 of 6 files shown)


davitol:
Testing on going:
- [x] remove non-whitelisted apps from navigation menu
- [ ] cosmetic removal of items in files view (side nav, buttons, tabs)
- [x] block /dav/comments if not whitelisted

PVince81:
While this might visually work, it's a hacky approach. Yes, there might be currently no better way due to the lack of extensibility of this bit of core.

The problems with it:

1) We're lucky that the navigation bar is in a dropdown. In the event where the whitelist call would take longer to go through, if a user opens the navigation dropdown, they'd see the icons there until the whitelist call hides them.

2) If a user knows the direct URL of an app, they'll still be able to access it.

@cortho or is there already any mechanism in place to prevent app access ?

A possible solution for 1) would be to modify core to provide an extension point where navigation items are loaded. By default it would just directly call the callback / resolve the promise, but apps like guests could plug into it and do the whitelist ajax call. But if 2) is still a problem then let's not do that.

A clean solution for 2) would be to have the backend allow apps to filter navigation items (through the INavigationManager) and also block specific apps.

PVince81:
Aha, I found this: https://github.com/owncloud/guests/pull/165/files#diff-e6640f8ed24baa82f7a645413ca040b6R83

Apparently this backend method is there to block all kind of app routes.

@davitol can you verify that blocking the calendar app will prevent opening the app's URL if you open the calendar app URL directly ? (copy-paste it before blacklisting)

cortho:
@PVince81 Yes, unfortunately @felixboehm and me are aware of this being a hack, still I witnessed, that there is some priority on this issue.

1. ack, but I don't see another solution yet
2. Direct access to the apps is prevented by a 403 header ( see https://github.com/owncloud/guests/blob/master/lib/appwhitelist.php#L54, which was introduced by @IljaN in the initial commit and I extended to consider URLs with /dav/comments as well. In all my tests this was working fine, so I don't think this is an issue.

PVince81:
@cortho sounds good.

After @davitol has verified that the URL for calendar is properly blocked (is this in the test plans?) I'm fine merging this 👍

davitol:
> @davitol can you verify that blocking the calendar app will prevent opening the app's URL if you open the calendar app URL directly ? (copy-paste it before blacklisting)

screen shot 2017-10-10 at 09 37 55

Issue #167, opened 2017-10-12
Disallow creating a guest if a user with the same email already exists


PVince81:
### Steps

1. In the users page create a user with the email "mail@example.com"
1. Go to the share panel for any file/folder
1. Type in "mail@example.com"
1. Select "mail@example.com (Guest)"

### Expected result
No "Guest" entry in the autocomplete results.
Clicking the result should not create anything.

### Actual result
"Guest" entry appears.
Clicking it will create a bogus duplicate user.

### Version
Latest master of the app.

PVince81:
PR here: https://github.com/owncloud/guests/pull/168

also cc @pmaier1 FYI

PR #168, opened 2017-10-12
Detect existing email/user on frontend and backend

PVince81:
Prevent creating users using an email that already exists on the
backend.

The frontend now discards the guest creation entry if there is at least
one exact match for the search terms, which could refer to the user id
or email address (the latter being handled by the autocomplete backend)

Fixes https://github.com/owncloud/guests/issues/167

Requires https://github.com/owncloud/core/pull/29223 from core to be able to detect exact user matches. Without this core PR the app will still work, but the "Guest" entry will be visible. Clicking it will however display an error message. I think this is acceptable for people still on 10.0.3.

### DIFF
--- a/controller/userscontroller.php
+++ b/controller/userscontroller.php
@@ -117,6 +117,13 @@ class UsersController extends Controller {
 			);
 		}
 
+		$users = $this->userManager->getByEmail($email);
+		if (!empty($users)) {
+			$errorMessages['email'] = (string)$this->l10n->t(
+				'A username with that email already exists.'
+			);
+		}
+
 		if (!empty($errorMessages)) {
 			return new DataResponse(
 				[
--- a/js/guestshare.js
+++ b/js/guestshare.js
@@ -102,10 +102,34 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 					}
 
 					// only allow guest creation entry if there is no exact match (by user id or email, decided by the server)
+					var provideGuestEntry = false;
+
 					if (xhrResult
 						&& xhrResult.ocs.meta.statuscode === 100
 						&& xhrResult.ocs.data.exact.users.length === 0
 					) {
+						provideGuestEntry = true;
+					}
+
+					// compatibility with OC <= 10.0.3 where xhrResult is not available
+					// here we always show the entry as we don't know about exact matches,
+					// and the backend might block the request if the guest is referring
+					// to an existing email address
+					if (!xhrResult) {
+						var lowerSearchTerm = searchTerm.toLowerCase();
+						if (!_.find(result, function(entry) {
+							if (entry && entry.value
+								&& entry.value.shareType === OC.Share.SHARE_TYPE_USER
+								&& entry.value.shareWith.toLowerCase() === lowerSearchTerm) {
+								return true;
+							}
+							return false;
+						})) {
+							provideGuestEntry = true;
+						}
+					}
+
+					if (provideGuestEntry) {
 						result.push({
 							label: t('core', 'Add {unknown} (guest)', {unknown: searchTerm}),
 							value: {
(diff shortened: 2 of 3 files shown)


PVince81:
To test:

- [x] TEST: existing local user who has an email address
- [x] TEST: existing local user whoses user id is the email address
- [ ] TEST: existing LDAP user with email address, but login/display name is different
- [ ] TEST: existing LDAP user with email address used also as login / user id
- [ ] TEST: existing LDAP user where email address is in a secondary search term field

@davitol can you test the LDAP cases above ? For the secondary search terms you might want to ask @tomneedham how to do it

cdamken:
[ ] TEST: guest user can be created with a non-duplicate *guest* email address

cdamken:
untitled

cdamken:
untitled

cdamken:
@PVince81 @cortho Now is not possible to create new users.

PVince81:
Works for me even when disabled federated sharing.

Would be good to check the JS console and config report in case some non-default setting were used with sharing that might affect this somehow.

cortho:
Just **double checked** the creation of new guest users, which was working as expected with current core master and this branch. @cdamken Could it be possible, that in your case the guests app was not enabled for any reason?

@PVince81 fyi

davitol:
* A current scenario is:
1 -You can share a file A ```guest@mail.com```  and it appears (guest) in the sharebox 
2 - You share another file B with ```guest@mail.com``` and (guest ) is no longer shown even when that guest user haven't accepted the mail yet. The mail receives both sharing mails from file A and B.
3 - When guest user accepts through file B mail, when he logs in. The file A is also included in his account

davitol:
> TEST: existing LDAP user with email address, but login/display name is different
> TEST: existing LDAP user with email address used also as login / user id

Tested and works

PVince81:
Let's not merge until we get clarification from @cdamken. I wonder if some option like "restrict sharing to own groups" or something is checked ?

PVince81:
@cdamken any news ? This important PR is blocked. Important because it fixes an issue that could cause trouble. Blocked because you found a case (need env info) which seem to break guest sharing completely.

davitol:
> TEST: existing LDAP user where email address is in a secondary search term field

@PVince81 userManager->find() by default searches for displayname, email, and their search terms
so email in their search terms wont have an effect. 

@cdamken any news or info about the apps enabled, whitelist apps, environment etc so we can move forward with this PR?

PVince81:
@felixboehm can you help here ?

PVince81:
Ok, turns out this PR doesn't work with OC 10.0.3 but only with OC 10.0.4...

I'll try and make it compatible, it was supposed to be compatible.

PVince81:
- [ ] TEST: all of the above with 10.0.3 🚫 FAIL

PVince81:
looks like I was too optimistic with the code and discarded my former workaround for 10.0.3... optimistic that people would upgrade the core version. Ok, I'll add back the workaround to make it compatible whenever the core changes are not there.

PVince81:
Ok, the last commit brings compatibility with 10.0.2 and 10.0.3.

I have tested the basic case with 10.0.2 and 10.0.3:
- the "Add guest" entry properly appears, always
- selecting "Add guest" for an existing email will show an error popup

Only with 10.0.4 the "Add guest" entry will disappear for existing email addresses.

PVince81:
@davitol can you retest, also with 10.0.3 ?

@cortho can you check the code additions ? it's basically the old code as fallback

davitol:
> @davitol can you retest, also with 10.0.3 ?

@PVince81 LGTM in 10.0.3. The guest user dropdown option is no longer available when the mail written has been already used by other guest user

cortho:
@PVince81 code looks fine to me

PVince81:
- [ ] requires https://github.com/owncloud/core/pull/29332 to be able to work correctly to deduplicate emails in a case insensitive manner

PVince81:
@davitol ok to merge then ? The core PR is independent and an additional fix.

davitol:
> @davitol ok to merge then ? The core PR is independent and an additional fix.

@PVince81 Yes, I think is fine for merging

PR #170, opened 2017-10-18
slightly more exact email check

### LINKED ISSUE #169: Offer 'add as guest' only when a valid mail address is entered
I think the guests app autocomplete should only offer adding a user as a guest when a valid mail address (x@y.tld) is entered to guide users. See screenshot.

screenshot from 2017-10-13 11-09-52

What is a bit odd is that you immediately get to see the warning when no internal user is found while typing (see second screenshot). Not sure if we can somehow avoid this. Ideas?

screenshot from 2017-10-13 11-11-53

Edit: Screenshots were taken with the unreleased, 'modified' guests app (no modal etc.)

@PVince81 @cortho

cortho:
Now the creation of guest users is offered when there is the @ character and at least one unicode or ascii character before and after the @.

fixes #169

### DIFF
--- a/js/guestshare.js
+++ b/js/guestshare.js
@@ -88,9 +88,13 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 
 		    return oldHandler.call(obj, search, function(result) {
 				var searchTerm = search.term.trim();
+				// before and after the @ sign we match at least one unicode char
+				// which is not in the range below or a word character
+				// (which is exempt in the unicode range)
+				var emailRegex = /^.*([^\u0000-\u007F]|\w).*@([^\u0000-\u007F]|\w)+.*$/;
 
 				// Add potential guests to the suggestions
-				if (searchTerm.search("@") !== -1) {
+				if (emailRegex.exec(searchTerm) !== null) {
 					// FIXME: will need some new hooks in core to be able to do this in a clean way
 					if (!result || !result.length) {
 						// no results, need to hack the message and still display something
@@ -124,7 +128,7 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 				response(result);
 		    });
 		};
-		
+
 		obj._onSelectRecipient = function (e, s) {
 			e.preventDefault();
 


js/guestshare.js
```diff
@@ -88,9 +88,10 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 
 		    return oldHandler.call(obj, search, function(result) {
 				var searchTerm = search.term.trim();
+				var emailRegex = /.*([^\u0000-\u007F]|\w).*@([^\u0000-\u007F]|\w)+.*/;
```
PVince81: not sure I understand, but from what I remember `^` used in the middle says "NOT", so are you excluding some unicode controls chars here ? But then you combine this with an OR operation `|`. So I'd read this as "no unicode control char OR an alphanumerical char". If this is what is intended, then I think we don't need `\w` because the negation of all unicode chars would already match any other chars, including alphanumerical ones.

I wonder if it wouldn't be better to just check:
- whether "@" is neither at the beginning nor the end of the string
- the string is at least 3 chars long
- there is only a single "@" sign

cortho: ^ and $ at the beginning and the end of a pattern are respectively meaning the beginning and the end of the string, ie: we match the entire string

PVince81: but in your pattern you have a `.*` at the beginning, so `^` isn't at the beginning.
Also, to mean "beginning" I'd expect the `^` to be outside of the square brackets.

PVince81: now we could also ask the question differently: are there any chars that we want to forbid inside the email address, given that the "@" is somewhere in the middle ? I don't think we should bother much about excluding chars here, so matching anything would be fine. (assuming that the backend properly validates the email afterwards before sending the email)

cortho: I was testing the regex again: IMHO this is quite a good basic test on email addresses. sth. like .@.  or string@ or \@string are not matched, but unicode chars are (except from the control characters) - good enough for UX. In the backend we're validating with the mailer's validate method: https://github.com/owncloud/guests/blob/ebc56266c87bdf4ffad8543e385c492a4b63d59c/controller/userscontroller.php#L108

Suggestion: I'll add the leading/trailing ^$ and leave the rest alone, we still can modify, when core or some lib provide a better test

js/guestshare.js
```diff
@@ -88,9 +88,10 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 
 		    return oldHandler.call(obj, search, function(result) {
 				var searchTerm = search.term.trim();
+				var emailRegex = /^.*([^\u0000-\u007F]|\w).*@([^\u0000-\u007F]|\w)+.*$/;
```
PVince81: this regexp still forbids unicode chars between 00 and 7F, is this intended ?

see https://stackoverflow.com/a/1763582

I think you should remove the `^` in the square brackets to avoid negation

PVince81: okay, then we need to talk about the `|` operator there. The logic here is "not a unicode char or a word". I think "not a unicode char" already matches **ANY** character that is not unicode, including non-word (`\w`) chars. Did you intend this "|" to be an **AND** operator actually, saying "a word char but not a unicode control char" ?

PVince81: spectacle c12215

PVince81: Just tested and you can see that it also matches brackets `(`

spectacle t12234

cortho: Yes, sure this is only a rudimentary check, I was assuming
* **email address validation is a core/3rd party lib affair, not in a single method in an app**
* we do have sound validation in the backend
* don't have a fancy regex which nobody can read
* only have a basic check, throwing out the most bad occurrences

maybe only checking for the @ was the best way (i think remote shares are handled the same way, so it would be at least consistent)

PVince81: @cortho yes, I agree that a simple check is alright. My worry is only that the current regexp itself is not self-consistent / does not make sense. It's a bit like writing `if ($a || $b)` knowing that `$a` is always true, in which case you can remove `$b` as it might confuse developers looking at it.

Let's then only check for the "@" but make sure that there are at least one character before, and two characters after to avoid patterns like "a@" to match too early ?

cortho: I wanted to match
(any character)(at least one non-control unicode character|ascii word)@(at least one non-control unicode character|ascii word)(any character)

[^\u0000-\u007F] matches ö but doesn't match a
\w matches a but doesn't match ö

cortho: @PVince81 whatever you prefer. I'm fine with the regex above, but if this causes headache for you, I can check for (at least one non-control unicode character|ascii word)@ as well

PVince81: Ok, I used a regexp tester and I found that "a" is indeed in the set of unicode chars between 00 and 7F, so it also excludes regular ASCII characters. So using this you're excluding more than just the control characters and punctuation. You also exclude regular alphabet letters and reinclude them with `\w`.

I'm fine to keep this now, please add a JS comment to clarify this and avoid future headaches. Thanks!

cortho: Of course, I'll add this.

cortho:
@PVince81 added comment

davitol:
Tested and works 👍

PR #192, opened 2018-02-26
Bump core max version to 10.1

phil-davis:
and adjust Travis matrix to run PHP 5.6+ against ``core`` ``stable10`` and PHP 7.1+ against ``master``

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -25,8 +25,7 @@ Furthermore guests are fully auditable with the [ownCloud Auditing application](
 	<licence>GPLv2</licence>
 	<author>ownCloud GmbH</author>
 	<dependencies>
-		<owncloud min-version="10.0" max-version="10.0" />
-		<php min-version="5.6" />
+		<owncloud min-version="10.0" max-version="10.1" />
 	</dependencies>
 	<category>collaboration</category>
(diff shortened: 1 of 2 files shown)


phil-davis:
It works. So that seems "a good thing" for now. 
The issue for the future will be that this has a ``composer.json`` and ``composer.lock`` - so the versions of dependencies are being locked to whatever is in ``composer.lock`` that has to have been generated on a PHP 5.6 system (since it needs to run on PHP from 5.6 to 7.2. There will be dependency updates that could be implemented for when running this app on top of ``core`` ``master`` with guaranteed PHP min 7.1.

appinfo/info.xml
```diff
@@ -25,7 +25,7 @@ Furthermore guests are fully auditable with the [ownCloud Auditing application](
 	<licence>GPLv2</licence>
 	<author>ownCloud GmbH</author>
 	<dependencies>
-		<owncloud min-version="10.0" max-version="10.0" />
+		<owncloud min-version="10.0" max-version="10.1" />
 		<php min-version="5.6" />
```
DeepDiver1975: this can go

phil-davis: gone

PR #200, opened 2018-04-01
Refactor and cleanup guests app. Add unit tests

mrow4a:
Goals of this PR (finally found time to push the code from local branch, happy easter :>) 

Generaly, the logic did not change, just refactored it a bit to be cleaner to understand and easier to unit-test and control in general

- [x] Remove from all controller classes the core logic, and use GuestsHandler, AppWhitelist and Mail classes instead. 
- [x] Clean up static functions mess from the classes
- [x] Clean up hooks so that legacy code is separated from new styled code (AppWhitelist, Mail, Hooks)
- [x] Remove very dangerous mechanism which allowed to dynamically set gid of guest group - use displayname instead (core needs adjustments to use group displaynames in UX?)
- [x] Confirmed that on stable10 app update is successfull
- [x] Add unit tests to newly created/moved classes
- [ ] Migration needed due to making group unique GID static (previously it was allowed that gid changes) - might need core adjustment to use displayname? 

@PVince81 @DeepDiver1975 @IljaN


mrow4a:
If https://github.com/owncloud/core/pull/29107 gets merged before this one, we need to check if update works all fine

appinfo/app.php
```diff
@@ -31,20 +32,30 @@ function() {
 	}
 );
 
-$config = \OC::$server->getConfig();
-$groupName = $config->getAppValue('guests', 'group', \OCA\Guests\GroupBackend::DEFAULT_NAME);
```
mrow4a: This might need migration probably, since we previously allowed to change group unique id dynamically

PVince81:
@mrow4a so this PR is just pure refactoring without any changes related to the central groups table ?

mrow4a:
Yes, and this Pr should make it easier and safer to integrate central groups. But disadvantage is that now only guest group displayname can change, and GID is fixed once created.

mrow4a:
Probably needs this first - https://github.com/owncloud/user_management/pull/28

PVince81:
there were many changes in the app and this would likely need to be redone in most part

for example user controller is gone

best would be to do this in smaller iterations

PR #202, opened 2018-04-14
delete unneeded mailhog url

phil-davis:
we don't need that mailhogUrl setting any-more
see owncloud/core#31079


individual-it:
superseded by #203

PR #209, opened 2018-05-30
Add drone

patrickjahns:
# Motivation

Use drone in our CI pipeline -> streamline automated testing

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,78 @@
+workspace:
+  base: /var/www/owncloud
+  path: apps/guests
+
+branches: [master, release*, release/*]
+
+pipeline:
+  install-server:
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -0,0 +1,84 @@
+workspace:
+  base: /var/www/owncloud
+  path: apps/guests
```
PVince81: at some point I guess you could use a variable name for app names for better file reusability ?

patrickjahns: yaml allows for anchors - would need to dig deeper if something like you mentioned is possible. But as usual - using these things makes it less "human readable" and sometimes copy+pasta is more readable than the most fancy constructs we can think of. sometimes the DRY principle is over exercised ;-)

PR #186, opened 2018-01-17
Add karma for JS tests

PVince81:
@DeepDiver1975 can you check travis.yml ? I have the feeling there is some kind of conflict between using env variables in the included matrix vs what is at the top of the file.

PR #210, opened 2018-05-30
Add karma for JS tests (in drone)

patrickjahns:
bootstrapping javascript test inside drone 

REF: https://github.com/owncloud/guests/pull/186


PVince81:
ok, so the test failure is because some of the code is already running complex logic in `$(document).ready()` instead of being triggered by tests. I'll restructure this to avoid triggering logic outside of tests.

PVince81:
@patrickjahns I had to modify the original source code to not boot the app. Moved the trigger for booting to "app.js" which isn't included in the code that the tests load.

To confirm the app still works I manually tested the following:
- [x] TEST: settings page still can load and save settings (whitelist) (for guests.js)
- [x] TEST: navigation entry is still hidden when not in whitelist (tested with activity app not whitelisted) (for navigation.js)
- [x] TEST: ability to share with guest, the "(guest)" entry appears in the autocomplete (for guestshare.js)

@patrickjahns anything else to add from your side ? The one dummy JS test pass now with Drone.

patrickjahns:
👍  - LGTM - we can squash (if wanted) and merge - or direct merge

note: i've not boostrapped karam for travis - will deprecate travis over the next days for `GUESTS`

PVince81:
:+1: for drone commit

Issue #198, opened 2018-03-21
Upload fails on files > 10 MB


voroyam:
If a guest user uploads a file that is bigger than 10 MB the upload fails without a helpful error message in the web UI or the ownCloud log

# Steps to reproduce:

- Create a share with a guest user
- Log in as guest user and upload a file in the shared folder
- Files < 10 MB get uploaded
- Files > 10 MB don't get uploaded

Tested with OC 10.0.3, 10.0.4, 10.0.7

Tested on docker Ubuntu with apache2 PHP 7 

enabled apps: guest app.

This is the only thing in the owncloud.log, but it seems not related to the problem, because it is also displayed if the upload is successful.

```
App::"filesystem",
"method":"MKCOL","url":"\/owncloud\/remote.php\/dav\/uploads\/voroyam%40gmail.com\/web-file-upload-566fa2790c670e81fb3048164457ba1b-1521636862684",
"message":"Storage wrapper 'oc_readonly' was not registered via the 'OC_Filesystem - preSetup' hook which could cause potential problems."}
App::"filesystem",
"method":"DELETE","url":"\/owncloud\/remote.php\/dav\/uploads\/voroyam%40gmail.com\/web-file-upload-566fa2790c670e81fb3048164457ba1b-1521636862684",
"message":"Storage wrapper 'oc_readonly' was not registered via the 'OC_Filesystem - preSetup' hook which could cause potential problems."}
App::"filesystem",
"method":"GET","url":"\/owncloud\/index.php\/apps\/files\/ajax\/getstoragestats.php?dir=%2Fforguests",
"message":"Storage wrapper 'oc_readonly' was not registered via the 'OC_Filesystem - preSetup' hook which could cause potential problems."}
```

here is the network tab taken during a failed upload

![network](https://user-images.githubusercontent.com/24212810/37713156-4a5f39f4-2d16-11e8-9fe4-3df920523a58.PNG)

@tomneedham @butonic @PVince81

cortho:
I guess when guests do not have a home some temporary files cannot be created. 

So This is not a guests app issue but a core/webdav issue. Suggestion: close here and reopen in core.

See my logs below.

**js errors**

`client.js?v=d5106cc…:246 MKCOL http://localhost/oc10/remote.php/dav/uploads/guest%40localhost/web-file-upload-5875dff62b65ed35356d174af1689b71-1521709720470 403 (Forbidden)`

[...]

PR #212, opened 2018-05-30
Test for chunking upload 

IljaN:
Test for #198 
Requires https://github.com/owncloud/core/pull/31596 to be merged

### DIFF
--- a/tests/acceptance/features/apiGuests/guests.feature
+++ b/tests/acceptance/features/apiGuests/guests.feature
@@ -55,6 +55,7 @@ Feature: Guests
     And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
     And user "guest@example.com" moves new chunk file with id "chunking-42" to "/tmp/myChunkedFile.txt" using the API
     Then as "guest@example.com" the file "/tmp/myChunkedFile.txt" should exist
+    And as "user0" the file "/tmp/myChunkedFile.txt" should exist
 
   @mailhog
   Scenario: A guest user can cancel a chunked upload
@@ -71,6 +72,7 @@ Feature: Guests
     And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
     And user "guest@example.com" cancels chunking-upload with id "chunking-42" using the API
     Then the HTTP status code should be "204"
+    And as "user0" the file "/tmp/myChunkedFile.txt" should not exist
 
   @mailhog
   Scenario: A guest user can upload a file and can reshare it
--- a/tests/acceptance/features/apiGuests/guests.feature
+++ b/tests/acceptance/features/apiGuests/guests.feature
@@ -41,6 +41,38 @@ Feature: Guests
     Then the HTTP status code should be "201"
 
   @mailhog
+  Scenario: A guest user can upload chunked files
+    Given as user "admin"
+    And user "user0" has been created
+    And user "admin" has created guest user "guest" with email "guest@example.com"
+    And the HTTP status code should be "201"
+    And user "user0" has created a folder "/tmp"
+    And user "user0" has shared folder "/tmp" with user "guest@example.com"
+    And guest user "guest" has registered
+    When user "guest@example.com" creates a new chunking upload with id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "1" with "AAAAA" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "2" with "BBBBB" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
+    And user "guest@example.com" moves new chunk file with id "chunking-42" to "/tmp/myChunkedFile.txt" using the API
+    Then as "guest@example.com" the file "/tmp/myChunkedFile.txt" should exist
+
+  @mailhog
+  Scenario: A guest user can cancel a chunked upload
+    Given as user "admin"
+    And user "user0" has been created
+    And user "admin" has created guest user "guest" with email "guest@example.com"
+    And the HTTP status code should be "201"
+    And user "user0" has created a folder "/tmp"
+    And user "user0" has shared folder "/tmp" with user "guest@example.com"
+    And guest user "guest" has registered
+    When user "guest@example.com" creates a new chunking upload with id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "1" with "AAAAA" to id "chunking-42" using the API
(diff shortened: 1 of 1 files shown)


patrickjahns:
@IljaN - please:
- add respective labels 
- assign yourself as being responsible when creating the PR
- add to milestone 

On the content - did you check that the behat tests will transform the upload to chunking? 

Checking core feature files for webdav - we explicitly chunk: https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdav/webdav-related-new-endpoint.feature#L464-L473


something to consider - instead of uploading a large file, create the file during the test run - so people do not have to check out unecessary large github repositories

PVince81:
also, as discussed in the core PR:
- [ ] add a test that does a DELETE on the uploads folder for cancellation, to confirm there is no permission issue

IljaN:
There seems to be no pre-defined step to delete files outside of /$user/files, should i create one? @individual-it, @phil-davis

Maybe an abstraction step for the chunking process should be provided which does the whole process without needing to do it manually like this:
```
  @mailhog
  Scenario: A guest user can upload chunked files
    Given as user "admin"
    And user "user0" has been created
    And user "admin" has created guest user "guest" with email "guest@example.com"
    And the HTTP status code should be "201"
    And user "user0" has created a folder "/tmp"
    And user "user0" has shared folder "/tmp" with user "guest@example.com"
    And guest user "guest" has registered
    When user "guest@example.com" creates a new chunking upload with id "chunking-42" using the API
    And user "guest@example.com" uploads new chunk file "1" with "AAAAA" to id "chunking-42" using the API
    And user "guest@example.com" uploads new chunk file "2" with "BBBBB" to id "chunking-42" using the API
    And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
    And user "guest@example.com" moves new chunk file with id "chunking-42" to "/tmp/myChunkedFile.txt" using the API
    Then as "guest@example.com" the file "/tmp/myChunkedFile.txt" should exist
    When user "guest@example.com" deletes folder "/uploads/chunking-42" using the API
    Then as "guest@example.com" the file "/uploads/chunking-42" should not exist
```

IljaN:
@PVince81 test added

phil-davis:
@IljaN yes, compressing that list of chunking steps would be nice for "routine" chunking tests - it can assume some stuff and do it all. If you want detailed control (upload weird numbers of chunks of different sizes out-of-order,...)  then you can write it out as individual steps.
I will make a PR to provide that.

tests/acceptance/features/apiGuests/guests.feature
```diff
@@ -40,6 +40,38 @@ Feature: Guests
...
+    And the HTTP status code should be "201"
+    And user "user0" has created a folder "/tmp"
+    And user "user0" has shared folder "/tmp" with user "guest@example.com"
+    And guest user "guest" has registered
+    When user "guest@example.com" creates a new chunking upload with id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "1" with "AAAAA" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "2" with "BBBBB" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
+    And user "guest@example.com" moves new chunk file with id "chunking-42" to "/tmp/myChunkedFile.txt" using the API
+    Then as "guest@example.com" the file "/tmp/myChunkedFile.txt" should exist
```
phil-davis: You can also add:
```
And as "user0" the file "/tmp/myChunkedFile.txt" should exist
```
just to be sure it ended up where it should.

tests/acceptance/features/apiGuests/guests.feature
```diff
@@ -40,6 +40,38 @@ Feature: Guests
...
+    And the HTTP status code should be "201"
+    And user "user0" has created a folder "/tmp"
+    And user "user0" has shared folder "/tmp" with user "guest@example.com"
+    And guest user "guest" has registered
+    When user "guest@example.com" creates a new chunking upload with id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "1" with "AAAAA" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "2" with "BBBBB" to id "chunking-42" using the API
+    And user "guest@example.com" uploads new chunk file "3" with "CCCCC" to id "chunking-42" using the API
+    And user "guest@example.com" cancels chunking-upload with id "chunking-42" using the API
+    Then the HTTP status code should be "204"
```
phil-davis: If you are feeling like really making sure, you can say:
```
And as "user0" the file "/tmp/myChunkedFile.txt" should not exist
```
to be sure the file did not arrive. (Make sure the HTTP status did not lie)
(it seems odd to me that the cancel of chunking is status ``204`` - but I guess that indicates the cancel was successful)

patrickjahns:
@IljaN 
restarted - even though it is now green for `stable10` - these errors strike me - cc @PVince81 
```
[Fri Jun  1 23:26:33 2018] Could not get node for path: "uploads/guest@example.com/chunking-42/.file" : File with name //chunking-42 could not be located
```

PVince81:
first warning is expected, needs fixing in core at some point.

second one is strange... I'll have a look

PVince81:
- [ ] requires https://github.com/owncloud/core/pull/31618 for master

patrickjahns:
@PVince81 - missing master as the required tests step implementation is in core

PVince81:
@patrickjahns I'm not seeing the second error from your comment https://github.com/owncloud/guests/pull/212#issuecomment-394034965 when running tests locally. Was it from a specific drone run ?

PVince81:
@patrickjahns ok I see the error on Drone: https://drone.owncloud.com/owncloud/guests/18/43 line 190.

Since Drone runs with the daily tarball I should try with that one locally as well.

PVince81:
@patrickjahns I managed to reproduce this, I wasn't looking properly. Using a search for ".file" helped me find it in the local log.

It's a core issue: https://github.com/owncloud/core/issues/31631

Some of our Sabre plugins is trying to clean up custom properties and tries to get the node after it was already moved, so it doesn't find it. Nothing to worry about in the context of uploads, but I mentioned in the ticket that we should check whether a real issue might exist with regular files and custom properties.

PVince81:
added more verification steps as advised by @phil-davis 

master got the acceptance test step so I expect everything to be green on all branches now

phil-davis:
Looks good, assuming CI passes.

PVince81:
daily-master-qa... need to wait for tomorrow has it will contain the fix. stay tuned

PVince81:
restarted failed builds, expecting success with today's daily-master-qa

PR #126, opened 2017-07-21
add guestsfilelist

cortho:
for #87, functionality as in files_sharing. I hope I didn't miss something

who could review this?

PR #214, opened 2018-05-30
Add filelist section to view guest shares

### LINKED ISSUE #87: Non-admin user cannot see invited guest user list.
For a "normal" user it should be possible to
 * see which users he invited
 * see if the user have already registered
 * resend an invitation (Currently: When the passwort reset token is expired, the only way to "resend" is to ask the admin to delete the guest user and then reinvite and re-share. )
 * to delete guest users he or she invited

PVince81:
Resend of https://github.com/owncloud/guests/pull/126 on a local branch.

Fixes #87 

- [x] restructure app.js to be able to unit test
- [x] add JS unit tests for the list
- [ ] add PHP unit tests for the new controller


PVince81:
Added JS tests for the list. Note that the test don't cover the other JS classes that existed before, this is pending tech debt.

Pending: writing tests for the GuestShareController.

I had to adjust it because of core changes, considering that `formatShare` now has an extra argument. This somehow scares me a bit as the controller is extending a private core class and could break any time we touch such methods again...

tests/unit/phpunit.xml
```diff
@@ -0,0 +1,22 @@
...
+	</testsuite>
+	<!-- filters for code coverage -->
+	<filter>
+		<whitelist>
+			<directory suffix=".php">../../appinfo</directory>
+			<directory suffix=".php">../../lib</directory>
+			<directory suffix=".php">../../settings</directory>
+			<directory suffix=".php">../../controller</directory>
+		</whitelist>
+	</filter>
```
patrickjahns: I'd recommend to add a `output` directive for the coverage report - makes life easier for running and collecting code coverage from one location.

example https://github.com/owncloud/configreport/blob/master/tests/unit/phpunit.xml#L20-L23

PVince81: I stole this from https://github.com/owncloud/guests/pull/200 :see_no_evil: 

noted, I'll add it next time I touch this PR

PVince81:
- [ ] add "output" and coverage thingy

tests/unit/phpunit.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<phpunit bootstrap="bootstrap.php"
+		 strict="true"
+		 verbose="true"
+		 timeoutForSmallTests="900"
+		 timeoutForMediumTests="900"
+		 timeoutForLargeTests="900"
+		 beStrictAboutTestsThatDoNotTestAnything="true"
+		>
+	<testsuite name='ownCloud - Guests App Tests'>
```
patrickjahns: recomended to keep it short - so it can be called via `phpunit --testsuite mytestsuite`

patrickjahns:
if wanted - we can now split up the tests into more parallel jobs:
- javascript
- phpunit
- acceptance

PVince81:
@patrickjahns should we then merge this first and write PHP unit tests for the controller separately ?

note that the JS unit tests are mostly copy-pasted from the files_sharing app which has a similar file list view, and adjusted for endpoint URLs

PVince81:
rebased and solved conflict

PVince81:
will likely take 0.5md more to finish the unit tests

PVince81:
rebased and adjusted

somehow still don't like much that we need an extra controller partly copied from core

DeepDiver1975:
will never be finished -> close

Issue #174, opened 2017-10-24
Guests cannot change email (should they even be allowed to?)


svengit:
Hi,

right now guest users can change their email address in the settings menu. Confirmation mail is send and after clicking the link a small message 'Email changed successfully' is displayed and a 'email adress changed'-Email is send out.
Yet, the email is still the same and the log shows an error for the whole action (I don't quite understand why the interface shows success and the log shows errors, but that's another topic):

[...]

PR #224, opened 2018-07-03
Guest users shouldn't be able to create other guest users #114

### LINKED ISSUE #114: Guests users can create other guests users
With the currents Guests app approach . A guest user is able to create another guest user. Is this the intended behaviour? Have we considered the impact of this decision about creating guest users in this way? @pmaier1 FYI @cortho 

Impl concept here: https://github.com/owncloud/guests/issues/114#issuecomment-341802055

IljaN:
Tasks from https://github.com/owncloud/guests/issues/114#issuecomment-399955795
- [x] frontend: do not load JS file that extends the autocomplete with a guest entry
- [x] backend: guest creation endpoint must return 403

Fixes #114

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -24,31 +24,8 @@
  */
 
 $app = new \OCA\Guests\AppInfo\Application();
-$app->registerListeners();
-
-$config = \OC::$server->getConfig();
-$groupName = $config->getAppValue('guests', 'group', \OCA\Guests\GroupBackend::DEFAULT_NAME);
-
-$groupBackend = new \OCA\Guests\GroupBackend($groupName);
-\OC::$server->getGroupManager()->addBackend($groupBackend);
-
-$user = \OC::$server->getUserSession()->getUser();
-
-if ($user) {
-	// if the whitelist is used
-	if ($config->getAppValue('guests', 'usewhitelist', 'true') === 'true') {
-		\OCP\Util::connectHook('OC_Filesystem', 'preSetup', '\OCA\Guests\AppWhitelist', 'preSetup');
-		// apply whitelist to navigation if guest user
-		if ($groupBackend->inGroup($user->getUID(), $groupName)) {
-			\OCP\Util::addScript('guests', 'navigation');
-		}
-	}
-
-	// hide email change field via css for learned guests
-	if ($user->getBackendClassName() === 'Guests') {
-		\OCP\Util::addStyle('guests', 'personal');
-	}
-}
+$groupBackend = $app->registerBackend();
+$app->registerListeners($groupBackend);
 
 // this will initialize the
 \OCP\Util::addScript('guests', 'app');
--- a/controller/userscontroller.php
+++ b/controller/userscontroller.php
@@ -29,6 +29,7 @@ use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\IUserManager;
+use OCP\IUserSession;
 use OCP\Mail\IMailer;
 use OCP\Security\ISecureRandom;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
@@ -56,6 +57,14 @@ class UsersController extends Controller {
 	 * @var ISecureRandom
 	 */
 	private $secureRandom;
+	/**
+	 * @var EventDispatcherInterface
+	 */
+	private $eventDispatcher;
+	/**
+	 * @var IUserSession
+	 */
+	private $currentUser;
 
 	/**
 	 * UsersController constructor.
@@ -68,15 +77,18 @@ class UsersController extends Controller {
 	 * @param IMailer $mailer
 	 * @param ISecureRandom $secureRandom
 	 * @param EventDispatcherInterface $eventDispatcher
+	 * @param IUserSession $currentUser
 	 */
-	public function __construct($appName,
-								IRequest $request,
-								IUserManager $userManager,
-								IL10N $l10n,
-								IConfig $config,
-								IMailer $mailer,
-								ISecureRandom $secureRandom,
-								EventDispatcherInterface $eventDispatcher
+	public function __construct(
+		$appName,
+		IRequest $request,
+		IUserManager $userManager,
+		IL10N $l10n,
+		IConfig $config,
+		IMailer $mailer,
+		ISecureRandom $secureRandom,
+		EventDispatcherInterface $eventDispatcher,
+		IUserSession $currentUser
 	) {
 		parent::__construct($appName, $request);
 
@@ -86,6 +98,7 @@ class UsersController extends Controller {
 		$this->mailer = $mailer;
 		$this->secureRandom = $secureRandom;
 		$this->eventDispatcher = $eventDispatcher;
+		$this->currentUser = $currentUser;
 	}
 
 	/**
@@ -93,9 +106,9 @@ class UsersController extends Controller {
 	 * @NoCSRFRequired
 	 * @NoAdminRequired
 	 *
-	 * @param $username
-	 * @param $email
-	 * @param $displayName
+	 * @param string $email
+	 * @param string $displayName
+	 *
 	 * @return DataResponse
 	 */
 	public function create($email, $displayName) {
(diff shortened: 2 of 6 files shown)


IljaN:
Please disregard review request, I found a issue

IljaN:
Now ready for review

controller/userscontroller.php
```diff
@@ -131,6 +142,22 @@ public function create($email, $displayName) {
...
+		$uid = $this->currentUser->getUser()->getUID();
+		$isGuest = (bool) $this->config->getUserValue(
+			$uid, 'owncloud', 'isGuest', false
+		);
+
+		if ($isGuest) {
+			return new DataResponse(
+				[
+					'message' => (string)$this->l10n->t(
+						'A guest user can`t create other guest users.'
```
patrickjahns: Do we translate API responses? ( I assume this error is never shown in the ui )

Additionally - wonder about the wording - a simple `Creation of guest users not allowed` should be more generic and hint less on the implementation detail

PVince81: A quick grep of "DataResponse" in core shows that we do translate responses as some are displayed as-is in the UI.

Agreed on changing the wording. In general I expect this message to never be visible as the endpoint will not be triggered from the UI. It could only in a scenario where a regular user currently logged in concurrently get converted to a guest user (<= which is currently not possible yet)

@IljaN please fix your keyboard to output proper quotes `'` instead of backticks

appinfo/app.php
```diff
@@ -50,6 +42,20 @@ function () {
 		}
 	}
 
+	$uid = $user->getUID();
+	$isGuest = (bool) $config->getUserValue(
```
PVince81: I suggest we move this into the listener.

Reason is to avoid having too much login in app.php. There is already too much here.
Keeping in mind that app.php is executed at the time an app is loaded, and in some rare cases some services aren't initialized yet.

In the future we'd like to get rid of app.php altogether if possible.

IljaN: @PVince81 which listener would you suggest?

PVince81: simply move it into the loadAdditionScripts one that already exists below

controller/userscontroller.php
IljaN: Will change it but I don`t get the reasoning here: Shouldn't an error message be always to the point and descriptive as possible regardless of context? Maybe it will leak out to the UI trough a bug or appear in some log a admin will see. It also imho also helps from code readability standpoint as you parse trough the different error conditions in the code.

PVince81: Fine leaving it as is, well, just fix the quote ;-)

patrickjahns: Please check with @settermjd or @phil-davis about wording after all 

> It imho also helps from code readability standpoint as you parse trough the different error conditions in the code as in that you can read the intent of the condition in plain englisch and compare it to the implementation.

I'd say this part should be covered by tests and less by error messages / log messages. 
Which reminds me - we do not have any unit tests for this added functionality :'(

PVince81: indeed. the solution here is to move the block to a method in the `Application` class. then rewire the hook to call that method

then write unit test for the method as app.php is not testable

PVince81:
@IljaN can you finish this ?

PVince81:
@IljaN please finish this, it will likely not take more than 15 mins or so

IljaN:
@owncluders rebase

IljaN:
@ownclouders rebase

Makefile
```diff
@@ -118,7 +118,7 @@ test-php-codecheck:
 
 .PHONY: test-php-style
 test-php-style: $(composer_dev_deps)
-	$(composer_deps)/bin/php-cs-fixer fix -v --diff --diff-format udiff --dry-run --allow-risky yes
+	$(composer_deps)/bin/php-cs-fixer fix -v --diff --diff-format udiff --allow-risky yes
```
PVince81: keep dry-run ?

IljaN: Woops sorry accidentaly commited, fixing

PVince81:
conflicts. @VicDeo has refactored the loadAdditionalScripts into the Application class in another PR

VicDeo:
fasten the belts, hijacking in progress

VicDeo:
rebased

lib/AppInfo/Application.php
```diff
@@ -68,27 +69,74 @@ function ($c) use ($server) {
...
+			\OCP\Util::connectHook('OC_Filesystem', 'preSetup', '\OCA\Guests\AppWhitelist', 'preSetup');
+			// apply whitelist to navigation if guest user
+			if ($isGuest) {
+				\OCP\Util::addScript(self::APP_NAME, 'navigation');
 			}
+		}
+
+		// hide email change field via css for learned guests
+		if ($isGuest) {
+			\OCP\Util::addStyle(self::APP_NAME, 'personal');
```
VicDeo: @PVince81 I noticed that 'css/personal.css' was deleted in #78 
I can restore this stylesheet but  hiding with CSS is not the best option IMO

PVince81: If you have a better quick solution please do.
Normally user backends can advertise a capability to allow/deny email change but it might not be possible for guests as they are regular DB users.

If not, restore the CSS and make a ticket

PVince81: hmm also see https://github.com/owncloud/guests/issues/174

PVince81: let's leave it out in this PR and address again as part of #174 to avoid mixing up topics

PVince81:
wait for https://github.com/owncloud/guests/pull/249 then rebase

phil-davis:
CI here is failing with:
```
    And guest user "guest" has registered                                                                                                  # GuestsContext::guestUserRegisters()
      Notice: Undefined offset: 0 in /var/www/owncloud/apps/guests/tests/acceptance/features/bootstrap/GuestsContext.php line 259
```
e.g. https://drone.owncloud.com/owncloud/guests/241/24

Something is wrong with it finding the guest registration email that should be received in Mailhog.

As above, I am reworking some stuff in #249 and am in that area of the acceptance test code.

VicDeo:
The failure is related to the assumption that guest can upload own files that is non correct according to my tests.

phil-davis:
#249 has been merged. Maybe it is worth rebasing this PR just to see what tests are still failing.

PVince81:
rebased to get master's max-version change

PVince81:
```
--- Failed scenarios:

    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:32
    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:44
    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:61
    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:78
    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:94
    /var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:130
/var/www/owncloud/apps/guests/tests/acceptance/features/apiGuests/guests.feature:162
```

PVince81:
@VicDeo can you finish this ? we plan a new guest release for another fix so would be good to have this one in as well, unless this takes more than half a day to fix

VicDeo:
@PVince81 something with the test env?
```
# GuestsContext::userCreatesAGuestUser()
      │ {"message":"User successfully created"}
      │ 
    And the HTTP status code should be "201"                                                                                               # FeatureContext::theHTTPStatusCodeShouldBe()
    And user "user0" has created a folder "/tmp"                                                                                           # FeatureContext::userCreatesAFolder()
    And user "user0" has shared folder "/tmp" with user "guest@example.com"                                                                # FeatureContext::userSharesFileWithUserUsingTheSharingApi()
    And guest user "guest" has registered                                                                                                  # GuestsContext::guestUserRegisters()
      No guest registration email was found on the email server
      Failed asserting that 0 is greater than 0.
```

from here: https://github.com/owncloud/guests/pull/265

VicDeo:
@PVince81 ok, the reason is in tests themself... somehow OC session has no user when they are being executed. 
My last commit clearly shows it. That's also the reason why I can't reproduce the failure manually.

PVince81:
hmmm weird, but without the refactoring they used to pass ?

VicDeo:
@PVince81 the hooks are registered unconditionally in master:
https://github.com/owncloud/guests/blob/master/appinfo/app.php#L26-L27
https://github.com/owncloud/guests/blob/master/lib/AppInfo/Application.php#L74

VicDeo:
@PVince81 ok, tests are passing

PR #230, opened 2018-08-09
Align acceptance test code with core

phil-davis:
1) Use ``ocPath`` that will go through the testing app
2) Use the acceptance test ```run.sh`` common script from core, and remove the local ``run.sh``
3) Add step ``userUploadsFileFromGuestsDataFolderTo`` to hide the gory details of how to reach the upload data files used by guest tests.
4) Init some capabilities in ``resetAppConfigs()`` - this keeps the ``AfterScenario`` code happy, which wants to check and restore capabilities to their previous state.


tests/acceptance/features/bootstrap/GuestsContext.php
```diff
@@ -48,10 +49,72 @@ class GuestsContext implements Context, SnippetAcceptingContext {
...
+	 */
+	private $relativePathToTestDataFolder = '../../apps/guests/tests/acceptance/data/';
+
+	/**
+	 * disable CSRF
+	 *
+	 * @throws Exception
+	 * @return string the previous setting of csrf.disabled
+	 */
+	private function disableCSRFFromGuestsScenario() {
```
phil-davis: I gave these methods unique names, because I suspect that we should provide them in core (part of ``BasicStructure``). So I want to be able to use the names ``disableCSRF`` and ``setCSRFDotDisabled`` in ``BasicStructure`` and not have it immediately create a function name duplication here.

PR #234, opened 2018-08-28
add PHP 7.2 for objectstore tests

davitol:
add PHP 7.2 for objectstore tests


patrickjahns:
shouldn't be required in this case - as there "should" be no difference between testing guests with objectstore on php7.1. Objectstore app itself should be tested with php7-7.2 compatability

davitol:
Let's close the ticket if we consider it make us gain anything

PR #254, opened 2018-10-23
update readme file that contains the white list

davitol:
There are more apps added in the guests whitelist due to PR: https://github.com/owncloud/guests/pull/227


PVince81:
mind sending to the release branch instead as part of QA ?

PR #255, opened 2018-10-24
Assert email found

phil-davis:
during acceptance test guest registration process, so that the problem is easier to see.

Otherwise the acceptance test error is like:
```
    And guest user "guest" has registered                                                                                                  # GuestsContext::guestUserRegisters()
      Notice: Undefined offset: 0 in /var/www/owncloud/apps/guests/tests/acceptance/features/bootstrap/GuestsContext.php line 261
```
which is not helpful.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -25,7 +25,7 @@ Furthermore guests are fully auditable with the [ownCloud Auditing application](
 	<licence>GPLv2</licence>
 	<author>ownCloud GmbH</author>
 	<dependencies>
-		<owncloud min-version="10.0.8" max-version="10.1" />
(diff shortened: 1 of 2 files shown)


phil-davis:
@PVince81 @DeepDiver1975 the 2nd commit here, bump core max-version to 11.0, is needed to make CI willing to run against core master tarball. It does pass, so I guess that guests app is still working with core master?

What do we do to get other changes merged, like the first commit here?
(because the max-version bump is needed to "trick" CI into passing)

individual-it:
code changes look good to me, what to do with the version?

phil-davis:
As discussed, we bump core max-version in apps to 11.0.0.0 - which I have now done here.

PR #262, opened 2018-10-30
Bump core max-version v11.0.0.0

phil-davis:
Bump up to max-version 11.0.0.0 again so that CI will run against core master.


phil-davis:
Note: this puts guests app back the same way as all the other apps that have max-version 11.0.0.0. That makes the "ownCloud universe" consistent. We have to improve towards the semver solution in the coming days, and for that all apps will have something happen to max-version again anyway.

PR #263, opened 2018-10-30
Bump core max-version v11.0.0-alpha.1

phil-davis:
Testing


phil-davis:
This does not work because PHP ``version_compare()`` sees that the current core master version ``11.0.0.0`` comes after ``11.0.0.alpha``

So we can't mention ``alpha`` in app ``info.xml`` ``max-version`` until core is also able to do that.

Issue #236, opened 2018-09-14
user autofocus can be null


PVince81:
Scenario is not clear, this was observed in a log:
```
ErrorException: Undefined index: user_autofocus
#11 guests/templates/form.password.php(42): handleError
#10 guests/templates/form.password.php(42): include
#9 /var/www/owncloud/lib/private/Template/Base.php(219): load
#8 /var/www/owncloud/lib/private/Template/Base.php(194): fetchPage
#7 /var/www/owncloud/lib/private/legacy/template.php(236): fetchPage
#6 /var/www/owncloud/lib/public/AppFramework/Http/TemplateResponse.php(149): render
#5 /var/www/owncloud/lib/private/AppFramework/Http/Dispatcher.php(108): dispatch
#4 /var/www/owncloud/lib/private/AppFramework/App.php(100): main
#3 /var/www/owncloud/lib/private/AppFramework/Routing/RouteActionHandler.php(46): __invoke
#2 /var/www/owncloud/lib/private/Route/Router.php(342): match
#1 /var/www/owncloud/lib/base.php(909): handleRequest
#0 /var/www/owncloud/index.php(54): null
```

mrlinux1:
I got the same error with a customer:

owncloud.log:
User opens password page for guests app:
 {"reqId":"W7yiEz@ssb1CklHMdYUtxAAAAAM","level":3,"time":"2018-10-09T14:41:55+02:00","remoteAddr":"XXX.XXX.XXX.XXX","user":"--","app":"PHP","method":"GET","url":"\/index.php\/apps\/guests\/register\/XXXXXXXXXXXXXXXXXXXXXX","message":"Undefined index: user_autofocus at \/var\/www\/owncloud\/apps\/guests\/templates\/form.password.php#42"}

apache.log
User clicks submit button:

 Benutzer bestätigt submit button der Form:
XXX.XXX.XXX.XXX - "" [09/Oct/2018:14:43:33 +0200] "GET /index.php/apps/guests/register HTTP/1.1" 405 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15”

owncloud 10.0.9.5 with guests app 0.5.0

PVince81:
I've grepped the code and it seems it is **always** null as there is no code that seems to set it.

Either we deleted some important code or this is obsolete. Needs further research.

[...]

PR #264, opened 2018-11-02
[WIP] Remove user_autofocus expectation from paramters list

sharidas:
Add user_autofocus to the parameter list to fix
the error thrown in the logs.

### DIFF
--- a/templates/form.password.php
+++ b/templates/form.password.php
@@ -39,8 +39,7 @@
 		<p class="groupbottom<?php if (!empty($_['invalidpassword'])) { ?> shake<?php } ?>">
 			<input type="password" name="password" id="password" value=""
 				placeholder="<?php p($l->t('Password')); ?>"
-				<?php p($_['user_autofocus'] ? '' : 'autofocus'); ?>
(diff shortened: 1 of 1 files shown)


sharidas:
fixes https://github.com/owncloud/guests/issues/236

sharidas:
As of now I have tried to get rid of the error from log. And this code is WIP. Below are the steps which need to be done.

- [ ] analyze why this change was missing and compare with old version 10.0.8

PVince81:
did you grep for "user_autofocus" in core ? maybe that template was copied from core and had it.

sharidas:
Yes, I found the `user_autofocus` in the core at https://github.com/owncloud/core/blob/master/core/templates/login.php#L48 and https://github.com/owncloud/core/blob/master/core/templates/login.php#L58

This was the first step I did. Found it at login.php and logincontroller. And then made a WIP PR. @PVince81

sharidas:
Indeed the `user_autofocus` is not required here. The user name field is already filled with the email id. The only field which needs user attention or required to be filled is the password. So we can remove this safely and autofocus on the password by default. I have updated the pr.

Issue #111, opened 2017-07-18
The email provided for a guest user should be checked that has a right email address format


davitol:
Using PR https://github.com/owncloud/guests/pull/109

![screen shot 2017-07-18 at 10 49 22](https://user-images.githubusercontent.com/10596560/28308580-ede4a712-6ba6-11e7-84fb-fbb0040ed86a.png)

@cortho

cortho:
Basically this is a correct email address, same as user@localhost. In addition to that, there is a check for valid addresses at https://github.com/cortho/guests/blob/new_registration/controller/userscontroller.php#L103

This still is an issue?

davitol:
What about guest@owncloud.com.com ? It seems to be accepted. As said it is only an enhancement.

cortho:
Even this is a valid email address since the domain com.com is registered.

davitol:
👍 OK Let's close it

PR #274, opened 2018-11-27
Fix the email regular expression

sharidas:

## Description

This change improves the validation of the email address for guest user creation by sharing the file/folder. Before this change was applied, strings like `a@f` were considered as valid. Which is not correct behaviour. This change includes regular expression which helps to validate the email address, similar to the email address validation done for public link share.

## Motivation and Context

Improve the email address validation regular expression.

## How Has This Been Tested?

share public link with email
1. core master branch
   1. emails that are acceptable in the text box
      - `a@b.cc`
      - `a@b.c.dd`
      - `a@b.c.d.ee`
      - `a@b.c.d.e.ff` etc
   2. emaisl that are not acceptable in the text box
      - `a@b`
      - `a@b.c`
      - `a@b.c.d`

2. core stable10 branch
   1. emails that are acceptable in the text box
      - `a@b.cc`
      - `a@b.c.dd`
      - `a@b.c.d.ee`
      - `a@b.c.d.e.ff` etc
   2. emaisl that are not acceptable in the text box
      - `a@b`
      - `a@b.c`
      - `a@b.c.d`

Guest app mail address validation in the `sharing` -> `User and Groups` of a file/folder:
   1. emails that are acceptable in the text box
      - `a@b.cc`
      - `a@b.c.dd`
      - `a@b.c.d.ee`
      - `a@b.c.d.e.ff` etc
   2. emails that are not acceptable in the text box
      - `a@b`
      - `a@b.c`
      - `a@b.c.d`

## Types of changes

## Checklist:


cortho:
I'm not sure about this change
* the new code does not support unicode characters
* a valid RFC-email like myuser@local-server is not accepted more (we have use cases for such servers here)
* what does it fix, why should the regex be changed at all?
* is there an issue ticket for this PR?

PVince81:
We have an internal ticket for this.

If this would remove support for unicode characters then we should reconsider.
Take the best of both regexp between guests app and public link share and apply it to both.

@pmaier1  see previous comment

PVince81:
The main issue is preventing email addresses without domain like "name@gmail".
The idea was to reuse the public link dialog regexp in guests app.

cortho:
Yes, unicode is broken with this change.

Regarding the domains: user@myhost is a valid email-address when I set up a host with the name "myhost". This could break some internal/vpn environments, where I know such addressing is used for guests. I'm not convinced, we should reject RFC-valid addresses just for convenience.

We had something like this before https://github.com/owncloud/guests/issues/111

sharidas:
@cortho I will fix the unicode change

pmaier1:
> This could break some internal/vpn environments, where I know such addressing is used for guests.

Hmm, this worries me, of course. Are you sure that there is a real-world use case for guest creation via local mail servers?

> I'm not convinced, we should reject RFC-valid addresses just for convenience.

This request for change is actually not for convenience purposes but to prevent confusion and better guide users to use real-world mail addresses. We have the same validation in place for sending public links via mail from the web interface. Still, if there **really** are requirements for guest user creation via local mail servers, we need to reconsider.

cortho:
> > This could break some internal/vpn environments, where I know such addressing is used for guests.
> 
> Hmm, this worries me, of course. Are you sure that there is a real-world use case for guest creation via local mail servers?

I have one setup here, but since I was reading the ticket, I can have changes done. When this is clearly communicated along the release, this gives the chance for other admins to act.

In an optimum case there would be an admin settings, so depending on one ownCloud's purpose admins could decide, which checks they prefer.

I don't remember by heart, but on the php side we are checking with the swift mailer with $this->mailer->validateMailAddress($email). So in theory the address could be seen valid by js but invalid by the mailer when the rules don't match.

Basically, both checks should return more or less the same.

> 
> > I'm not convinced, we should reject RFC-valid addresses just for convenience.
> 
> This request for change is actually not for convenience purposes but to prevent confusion and better guide users to use real-world mail addresses. We have the same validation in place for sending public links via mail from the web interface. Still, if there **really** are requirements for guest user creation via local mail servers, we need to reconsider.

It's not only about the top-level domain. What about servers like a.example.org and b.example.org, which is a quite common setup. This wouldn't validate with the current PR too.

sharidas:
- Updated the code with unicode support. Also tested with the steps for guest app in https://github.com/owncloud/guests/pull/274#issue-233893565

pmaier1:
> I have one setup here, but since I was reading the ticket, I can have changes done. When this is clearly communicated along the release, this gives the chance for other admins to act.

Do I understand correctly that you don't have objections for the change ultimately? :)

> What about servers like a.example.org and b.example.org

These need to work, of course, yes.

sharidas:
With this pr `a.example.org` does work
vokoscreen-2018-11-27_19-15-20

cortho:
@pmaier1 
> > I have one setup here, but since I was reading the ticket, I can have changes done. When this is clearly communicated along the release, this gives the chance for other admins to act.
> 
> Do I understand correctly that you don't have objections for the change ultimately? :)

Yes, when you came to the conclusion this is better for the majority of users this is something we can live with. Additionally, this would only affect the UI, so when posting to an API, I guess I could still use some other email format.

@sharidas 
oh, sorry, you are right, this was working, but müller@example.com wouldn't.
And before the regex was negating lower unicode characters (control chars) with ^ inside the brackets, now they are allowed.

sharidas:
@cortho 
> oh, sorry, you are right, this was working, but mü[ller@example.com](mailto:ller@example.com) wouldn't.
> And before the regex was negating lower unicode characters (control chars) with ^ inside the brackets, now they are allowed.

Now it works.

sharidas:
I am having a thought now:
- [ ] Should we consider support of internationalized country code ( referring to : last minimum 2 chars for domain name ). Lets say  https://en.wikipedia.org/wiki/Internationalized_country_code_top-level_domain

pmaier1:
> Should we consider support of internationalized country code ( referring to : last minimum 2 chars for domain name ). Lets say https://en.wikipedia.org/wiki/Internationalized_country_code_top-level_domain

Wow, never heard of this. Seems to be approved by ICANN so if not much effort might be good to support it. Opinions?

cortho:
@pmaier1 Sure this must be supported (and was covered by the origin regex).

Maybe we're approaching this from a too technological point of view. I wonder, if not a very simple regex like
`/[^\s@]+@[^\s@]+\.[^\s@]{2,}$/`
would be enough

sharidas:
I have tested the regular expression:
`/^[^\s@]+@[^\s@]+\.[^\s@]{2,}$/` and it looks promising
testingregexp

I am going to update this PR with this regular expression.

PVince81:
@sharidas please move it into a utility function and add JS unit tests that confirm that all the cases above are covered and the excluded ones are properly excluded.

Here I still wonder if we should make that a core utility method, maybe "OC.validateEmail()" that we can reuse in both public link page and guests app. The core one might already have some unit tests.

sharidas:
As per the suggestion I have made following changes ( current status of this PR ):
- [x] Created PR in the core https://github.com/owncloud/core/pull/33695
- [x] Used the new change made https://github.com/owncloud/core/pull/33695, in this PR.

sharidas:
@cortho A couple of observation spotted with the regular expression `/^[^\s@]+@[^\s@]+\.[^\s@]{2,}$/` is allows email address mentioned below:

- Jill.Valentine@umbrella..com
- Jill.Valentine@um#rella.com

Shall we consider this to be ok? Or should we remove the usage of consecutive usage of `..` after `@` in email address? @PVince81 @pmaier1

sharidas:
I have made a minor change in the regex in core, to avoid `Jill.Valentine@umbrella..c` for validation. The regex validation does not look easy. Say for example it allows 
- Jill.Valentine@umbrella..com
- Jill.Valentine@umbrella.==
- Jill.Valentine@umbrella.__

Basically any no white space character after the last dot  :(

PVince81:
please also add `min-version=10.1` in info.xml as this is the core version that will contain the new API

sharidas:
> please also add `min-version=10.1` in info.xml as this is the core version that will contain the new API

After updating the PR, I got a query, the dependency section in info.xml before my update was:
```
<owncloud min-version="10.0.8" max-version="10" />
```
Since the max-version is "10", wouldn't "10.1" changes be applicable to this app?

PVince81:
@micbar please read the above discussion and see if it helps with your concerns

PVince81:
> > please also add `min-version=10.1` in info.xml as this is the core version that will contain the new API
> 
> After updating the PR, I got a query, the dependency section in info.xml before my update was:
> 
> ```
> <owncloud min-version="10.0.8" max-version="10" />
> ```
> 
> Since the max-version is "10", wouldn't "10.1" changes be applicable to this app?

@sharidas no, the min-version needs to apply first. So it will be OC >= 10.0.8 and OC < 11

micbar:
I am still wondering, why we would support Unicode characters. they are per RFC not allowed in the local part.
`müller@example.com` ist not a valid email address.

cortho:
@micbar Which RFC are you referring to? In https://tools.ietf.org/html/rfc6530 international email addresses such as δοκιμή@παράδειγμα.δοκιμή are discussed.

micbar:
> @micbar Which RFC are you referring to? In https://tools.ietf.org/html/rfc6530 international email addresses such as δοκιμή@παράδειγμα.δοκιμή are discussed.

I was referring to RFC5322. I am aware, that Unicode characters are discussed and available in the domain part, but I was unaware, that some providers support Unicode now.

PVince81:
ah of course, CI doesn't run until core has version 10.1...

davitol:
I have started testing this changelog item:



> core stable10 branch
> 
>     2. emails that are not acceptable in the text box
>        
>        * `a@b`

But that mail is accepted for sharing and creating a guest user using guests: 0.8.0 and oC 10.1.0.4. Is the list of mails accepted updated in the ticket?

(cc @sharidas @pmaier1 )

PVince81:
@sharidas can you recheck the latest rules ? it is possible that after the discussion in the PR that the ticket doesn't match the requirements any more. We need the final requirements for QA. Thanks

sharidas:
@davitol I am having a quite opposite experience, which matches with the point `a@b` email is not acceptable in the text box for creating guest user. Here is the screenshot captured from my instance: 
guests

Also with respect to the email validation in the public share here is the screen shot captured from my instance 
share

May I know if there is extra config added to config.php or anything else? In my case I have added email configuration for smtp and license key. Apart from that I have not modified any extra configuration.
The screeshot that I have shared is from scratch installation on stable10 branch + guest app 0.8.0.

davitol:
After a screensharing with @sharidas checking the bits from GH tag and from guest tarball. He was not able to reproduce it. So I have install the server from scratch and now I'm not able to reproduce it anymore. So maybe it was something wrong with my configuration or a misunderstanding testing. LGTM now

PR #284, opened 2018-12-04
Adjust 'has created a folder' test steps

phil-davis:


phil-davis:
I could not get the green result from drone to update here in GitHub.
Tries again in PR #285 and it worked.
Closing here.

PR #290, opened 2018-12-13
Fix share with guest_app endless loop, add unit tests

micbar:
## Description
Fix endless activity loop when sharing with "guest_app" group and add unit tests.

## Related Issue
https://github.com/owncloud/enterprise/issues/3041

## Motivation and Context
Discussion with @DeepDiver1975 . GroupInterface needs to be implemented correctly.
He requested to add unit tests.

## How Has This Been Tested?
- Manually
- Unit Tests

## Types of changes

## Checklist:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -91,6 +91,18 @@ pipeline:
       matrix:
         TEST_SUITE: owncloud-coding-standard
 
+  phpunit-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - COVERAGE=${COVERAGE}
+    commands:
+      - if [ -z "${COVERAGE}" ]; then make test-php-unit; fi
+      - if [ "${COVERAGE}" = "true" ]; then make test-php-unit-dbg; fi
+    when:
+      matrix:
+        TEST_SUITE: phpunit
+
   api-acceptance-tests:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
@@ -104,6 +116,18 @@ pipeline:
     when:
       matrix:
         TEST_SUITE: api-acceptance
+
+  codecov:
+    image: plugins/codecov:2
+    secrets: [codecov_token]
+    pull: true
+    files:
+      - '*.xml'
+    paths:
+      - tests/output
+    when:
+      matrix:
+        COVERAGE: true
 
   notify:
     image: plugins/slack:1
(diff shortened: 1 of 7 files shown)


phpunit.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<phpunit bootstrap="./tests/bootstrap.php"
+		 verbose="true" failOnWarning="true"
+		 timeoutForSmallTests="900"
+		 timeoutForMediumTests="900"
+		 timeoutForLargeTests="900"
+>
+	<testsuite name='guests-unit'>
```
phil-davis: just call the testsuite ``unit``
The "standard" ``Makefile`` PHP unit command runs the ``unit`` suite

tests/bootstrap.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
phil-davis: Can this go in ``tests/unit`` folder, like other apps?
Then it is clearly part of the PHP unit test infrastructure.

PVince81:
@micbar is it intended that this PR brings so much phpunit setup ? the PR description doesn't state so, so it isn't clear whether this was some kind of mistake.

if it was intended, please see @phil-davis's comments.

the fix itself looks fine to me

phil-davis:
@PVince81 Guests currently has no PHP unit tests. So the unfortunate sucker who writes the first PHP unit test gets to also insert the needed CI infrastructure!

micbar:
@DeepDiver1975 requested unit tests, as mentioned in the description.

tests/bootstrap.php
micbar: I can see in the other repos, that the file is in the top folder

micbar:
@phil-davis @PVince81 addressed your input.

micbar:
Currently, we run the unit tests against core master.
I
I assume we would need to run them against stable10 too?

Would be easy to me, to add this. 
Unfortunately, this PR has grown bigger than intended :-)

You decide.

tests/bootstrap.php
micbar: Tom told me in another project, to separate Unit Tests and Integration tests. We would need the bootstrap.php for both test suites?

How is the best practise here?

PVince81:
@micbar if composer.lock changes are not related to your fix, please submit these in a separate PR.

We should keep changes in a PR to the minimum related to the fix, unless really necessary.

phil-davis:
> @micbar if composer.lock changes are not related to your fix, please submit these in a separate PR.
> 
> We should keep changes in a PR to the minimum related to the fix, unless really necessary.

If you do not change ``composer.json`` then there should normally be no change to ``composer.lock`` - I guess you must have done ``composer update`` at some time, rather than just ``composer install``. Unless there is a reason that this PR needs the newer version bumps in ``composer.lock``, then remove it.

.drone.yml
```diff
@@ -91,6 +91,18 @@ pipeline:
...
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - COVERAGE=${COVERAGE}
+    commands:
+      - if [ -z "${COVERAGE}" ]; then make test-php-unit; fi
+      - if [ "${COVERAGE}" = "true" ]; then make test-php-unit-dbg; fi
+    when:
+      matrix:
+        TEST_SUITE: php-unit
```
phil-davis: Just call it ``phpunit`` to be consistent with what has already been done in all other apps.
That will save people some minor hassle in future when edited drone scripts in multiple repos.

.drone.yml
```diff
@@ -105,6 +117,18 @@ pipeline:
       matrix:
         TEST_SUITE: api-acceptance
 
+  codecov:
+    image: plugins/codecov:2
+    secrets: [codecov_token]
+    pull: true
+    files:
```
phil-davis: Your version will work fine.
But may as well make this section the same as in other apps:
```
  codecov:
    image: plugins/codecov:2
    secrets: [codecov_token]
    pull: true
    paths:
      - tests/output/*.xml
    when:
      matrix:
        COVERAGE: true
```

tests/bootstrap.php
phil-davis: Mostly we do not even need this file. ``phpunit.xml`` uses the one from core directly without change. So try that and see if it works. In other cases I have been putting or moving it into ``tests/unit`` in all the other repos.

.drone.yml
micbar: I tried that, didn't work. @patrickjahns recommended to do it this way.

.drone.yml
micbar: done

tests/bootstrap.php
micbar: ok, it works with the one from core. Removed it

micbar:
@phil-davis can you review again?

phil-davis:
@PVince81 @DeepDiver1975 are you guys happy with the actual code changes and unit test content?

micbar:
Last remark from my side. Unit tests are testing against OC master branch.

PR #305, opened 2019-02-04
Stick to PSR-4, increase coverage

VicDeo:
## Description
Fix for invisible shares has been moved to files_sharing
So I save some housekeeping work only.

## Types of changes

## Checklist:


PVince81:
@VicDeo for repairs you can specify repair step class in apps in info.xml

lib/Hooks.php
```diff
@@ -121,8 +127,24 @@ public function handlePostShare(IShare $share) {
 		);
 
 		try {
+			// Filesharing app fails to create non-default path for shared items
```
jvillafanez: Makes me wonder why it isn't fixed there but here.... I'm lacking info.

lib/Hooks.php
```diff
@@ -121,8 +127,24 @@ public function handlePostShare(IShare $share) {
...
+			$shareTarget = $share->getTarget();
+			if ($shareFolder !== '/' && \strpos($shareTarget, $shareFolder) === 0) {
+				$count = 1;
+				$newTarget = \str_replace(
+					$shareFolder,
+					'',
+					$shareTarget,
+					$count
+				);
+				$share->setTarget($newTarget);
```
jvillafanez: shouldn't we check that the newTarget doesn't exists, or is it handled already within the `updateShareForRecipient` function? I remember there is code around sharing to add the "(2)", "(3)", etc to prevent overlapping with possible existing files...

lib/Hooks.php
```diff
@@ -121,8 +127,24 @@ public function handlePostShare(IShare $share) {
 		);
 
 		try {
+			// Filesharing app fails to create non-default path for shared items
+			$shareFolder = $this->config->getSystemValue('share_folder', '/');
+			$shareTarget = $share->getTarget();
+			if ($shareFolder !== '/' && \strpos($shareTarget, $shareFolder) === 0) {
+				$count = 1;
+				$newTarget = \str_replace(
```
jvillafanez: I'd prefer one-line style here (no need to change anything if this is the only change)
```
$newTarget = \str_replace($shareFolder, '', $shareTarget, 1);
```

lib/Hooks.php
VicDeo: because sharing app knows nothing about guests while guests app operates on shares to create guests

VicDeo: Guests are using read-only storage. By default shares are created with file_target '/sharename'.
When `share_folder` is non-default shares are created as '/share_folder/sharename'.
As soon as guests are using read-only storage it's not possible to create `/share_folder` there.
So all shares exist but are not visible to guests.

lib/Hooks.php
VicDeo: `only variables could be passed by reference`. E.g. https://stackoverflow.com/questions/5842366/php-variables-can-be-passed-by-reference-in-str-replace

VicDeo: ... and as described there `count` is number of the replacements made and not the limit. So this part war rewritten using `substr_replace` instead

lib/Hooks.php
VicDeo: For shares it is handled by the mount manager.

VicDeo:
I left only PSR-4 and unit test for Hooks class here.
The fix has been moved to https://github.com/owncloud/core/pull/34388

PR #307, opened 2019-02-09
Trigger CI

phil-davis:
To see if the JS tests pass


phil-davis:
Problems in core have been fixed

PR #318, opened 2019-03-26
Fix high DB load on inGroup() and getUsersGroups() calls

micbar:
## Description
Don't call getUsersForUserValue in AllConfig if not necessary. All methods with a given user context should call getUserValue.

## Related Issue
Closes https://github.com/owncloud/enterprise/issues/3242

## Motivation and Context
The current GroupInterface is not performant

## Types of changes

## Checklist:

### DIFF
--- a/lib/GroupBackend.php
+++ b/lib/GroupBackend.php
@@ -131,7 +131,16 @@ class GroupBackend implements GroupInterface {
 	 * Checks whether the user is member of a group or not.
 	 */
 	public function inGroup($uid, $gid) {
-		return \in_array($uid, $this->getMembers()) && $gid === $this->groupName;
+		if ($gid !== $this->groupName) {
+			return false;
+		}
+		$isGuest = $this->config->getUserValue(
+			$uid,
+			'owncloud',
+			'isGuest',
+			'0'
+		);
+		return $isGuest === '1' ;
 	}
 
 	/**
@@ -146,7 +155,13 @@ class GroupBackend implements GroupInterface {
 	 * if the user exists at all.
 	 */
 	public function getUserGroups($uid) {
-		if (\in_array($uid, $this->getMembers())) {
+		$isGuest = $this->config->getUserValue(
+			$uid,
+			'owncloud',
+			'isGuest',
+			'0'
+		);
+		if ($isGuest === '1') {
 			return [$this->groupName];
 		}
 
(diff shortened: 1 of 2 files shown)


lib/GroupBackend.php
```diff
@@ -131,7 +131,13 @@ public function implementsActions($actions) {
 	 * Checks whether the user is member of a group or not.
 	 */
 	public function inGroup($uid, $gid) {
-		return \in_array($uid, $this->getMembers()) && $gid === $this->groupName;
+		$isGuest = $this->config->getUserValue(
+			$uid,
+			'owncloud',
+			'isGuest',
+			0
```
PVince81: should this be a string `'0'` ?

lib/GroupBackend.php
```diff
@@ -131,7 +131,13 @@ public function implementsActions($actions) {
...
+		);
+		return $isGuest === '1' && $gid === $this->groupName;
```
PVince81: how about doing the `$gid` check first to avoid calling `getUserValue` when not needed ?

same for the other changes

tests/unit/GroupBackendTest.php
```diff
@@ -66,9 +66,32 @@ public function setUp() {
...
+	 * @param string $arg2
+	 * @param string $arg3
+	 *
+	 * @return int|string
+	 */
+	public function getUserValue($userId, $arg2, $arg3) {
+		if ($arg2 !== 'owncloud' || $arg3 !== 'isGuest') {
+			return 0;
+		}
+		if (\in_array($userId, $this->members)) {
```
PVince81: let's use strict comparison by adding `true` as third param to every `in_array` call

tests/unit/GroupBackendTest.php
```diff
@@ -66,9 +66,32 @@ public function setUp() {
...
+			return '1';
+		}
+		return 0;
```
PVince81: '0' as string ?

lib/GroupBackend.php
micbar: good point, will do

lib/GroupBackend.php
micbar: configvalues are always strings, right? Then it should be a string

micbar: Correct, it is type "longtext"

tests/unit/GroupBackendTest.php
micbar: ok

tests/unit/GroupBackendTest.php
micbar: same as above

PR #320, opened 2019-03-26
Update yarn deps

PVince81:


PVince81:
this only affects JS unit testing so if CI passes, we're good

PR #339, opened 2019-06-20
guest app test against latest release tarball

HariBhandari07:
## Description

Currently we only run app tests against stable10 to make sure they will be compatible with the future OC core platform release.
This test guarantee that the ```guests``` app runs correctly against the last released core version.
## Related Issue

Part of https://github.com/owncloud/QA/issues/611
## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,7 +2,6 @@ workspace:
   base: /var/www/owncloud
   path: apps/guests
 
-
 pipeline:
   install-core:
     image: owncloudci/core
@@ -17,6 +16,18 @@ pipeline:
       matrix:
         NEED_CORE: true
 
+  install-testrunner:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - git clone -b stable10 --depth=1 https://github.com/owncloud/core.git /var/www/owncloud/testrunner
+      - cp -r /var/www/owncloud/apps/guests /var/www/owncloud/testrunner/apps/
+      - cd /var/www/owncloud/testrunner
+      - make
+    when:
+      matrix:
+        USE_RELEASE_TARBALL: true
+
   install-app:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -559,3 +596,32 @@ matrix:
       FIX_PERMISSIONS: true
       USE_EMAIL: true
       TEST_OBJECTSTORAGE: true
+
+      # Owncloud Tarball
+    - PHP_VERSION: 7.1
+      OC_VERSION: daily-stable10-qa
```
phil-davis: I guess this needs to point to `stable10` because drone will install the core `stable10` code to get the current acceptance test code, and the acceptance tests will run from `/var/www/owncloud/core/apps/guests`
But they will test against the system that is installed in `/var/www/owncloud/tarball`
Am I understanding correctly?

.drone.yml
```diff
@@ -559,3 +596,32 @@ matrix:
...
+      BEHAT_SUITE: apiGuests
+      DB_TYPE: mysql
+      DB_NAME: owncloud
+      NEED_SERVER: true
+      NEED_INSTALL_APP: true
+      USE_EMAIL: true
+      NEED_CORE: true
+      USE_TARBALL_SERVER: true
+      FIX_PERMISSIONS: true
+      TARBALL_VERSION: 10.2.0
```
phil-davis: This will need to be updated soon after a new core release.
We could say `latest` here, and it should get https://download.owncloud.org/community/owncloud-latest.tar.bz2
But that tarball currently has only version `10.0.10` core code???
I have asked if that tarball can be updated...

For now, `10.2.0` is the thing to write.

.drone.yml
```diff
@@ -27,7 +38,7 @@ pipeline:
       - php occ a:e testing
       - php occ a:l
       - php occ config:system:set trusted_domains 1 --value=owncloud
-      - php occ log:manage --level 0
+      - php occ log:manage --level 2
```
dpakach: I dont think this should be changed.

skshetry: We have reduced log level to 2 in most of the places. I think, this is okay.

DeepDiver1975:
I'd say the cla service is broken :man_shrugging:

PR #369, opened 2019-10-08
Use model addShare instead of share api while wrapping ShareDialogView

mrow4a:
Guest share is a wrapper over ShareDialogView and ShareModel. 

However, while adding share, share API was used instead of the wrapped class - this can cause issues like https://github.com/owncloud/richdocuments/issues/285

### DIFF
--- a/js/guestshare.js
+++ b/js/guestshare.js
@@ -47,38 +47,32 @@
 			};
 
 			$.ajax(xhrObject).done(function (xhr) {
-				self._addGuestShare();
+				var properties = {
+					shareType: 0,
+					shareWith: self.email.toLowerCase(),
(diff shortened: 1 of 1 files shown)


PVince81:
@micbar FYI would need a new guests app release

micbar:
@PVince81 is this important for secure view?

PVince81:
yes, likely. @mrow4a can confirm

mrow4a:
@micbar @PVince81 well, this will improve a little user experience using Secure View for guests (SecureView checkbox will appear immediately after creating share (instead of after disabling edit)

PVince81:
okay, so not that critical to release the app very soon I guess as there is a quick workaround

PR #371, opened 2019-10-08
Provide extended attributes when the event is emitted

sharidas:
Provide extended attributes when the event is emitted.

## Description

Provide extended attributes for the guest user. This could be used by other apps or even core. Say for example to remove the visibility in the settings page.

## Motivation and Context

Provide extended attributes of the guest user.

## Types of changes

## Checklist:

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -22,11 +22,14 @@
 
 namespace OCA\Guests\AppInfo;
 
+use OCA\Guests\AppWhitelist;
 use OCA\Guests\Hooks;
 use OCA\Guests\Mail;
 use OCP\AppFramework\App;
 use OCP\GroupInterface;
 use OCP\IConfig;
+use OCP\IServerContainer;
+use OCP\User\UserExtendedAttributesEvent;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/AppInfo/Application.php
```diff
@@ -68,6 +71,45 @@ function ($c) use ($server) {
...
+	 * Sets the app attributes for the user.
+	 * The guestUser, whitelistedApps are added only if they don't exist in the
+	 * attributes array. Also the attributes are updated only if the previous value
+	 * does not match with the current attribute value.
+	 *
+	 * @param IServerContainer $server
+	 * @param UserExtendedAttributesEvent $attributesEvent
+	 * @param $userAppAttributes
+	 */
+	private function addUserAppAttributes(IServerContainer $server, UserExtendedAttributesEvent $attributesEvent, $userAppAttributes) {
```
jvillafanez: This doesn't seem the right place for this code... any better alternative?

lib/AppInfo/Application.php
```diff
@@ -68,6 +71,45 @@ function ($c) use ($server) {
...
+		if ($isGuestUser !== '0') {
+			$appWhiteList = AppWhitelist::getWhitelist();
+
+			/**
+			 * Add whitelistedAppsForGuests attribute if its not present. There is a
+			 * case where the whitelistedAppsForGuests could be changed by the admin.
+			 * Under such circumstance, if the whiteListedApp is present in
+			 * userAppAttributes array we have to update it.
+			 */
+			if (!isset($userAppAttributes['whitelistedAppsForGuests']) ||
```
jvillafanez: If it isn't set you don't do anything. This is the same behaviour the `$event->setAttribute` has, so just set it directly.

phil-davis:
See #518 for the up-to-date rebased version to go with core 10.11

PR #376, opened 2019-10-25
Share the whitelisted apps data to core

sharidas:
This data should be available in the core,
so that under certain circumstances where core
can decide which apps to be shown for the user.

## Types of changes

## Checklist:

### DIFF
--- a/appinfo/Migrations/Version20191115135305.php
+++ b/appinfo/Migrations/Version20191115135305.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\guests\Migrations;
+
+use OCA\Guests\AppWhitelist;
+use OCA\Guests\GroupBackend;
+use OCP\App\IAppManager;
+use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\Migration\ISimpleMigration;
+use OCP\Migration\IOutput;
+
+/**
+ * Move the whitelisted data from the appconfig to appaccess, if its there.
+ * If no data is available then just update the appaccess with default whitelisted
+ * apps data.
+ */
+class Version20191115135305 implements ISimpleMigration {
+	/** @var IAppManager */
+	private $appManager;
+	/** @var IGroupManager */
+	private $groupManager;
+	/** @var IConfig */
+	private $config;
+
+	/**
+	 * Version20191115135305 constructor.
+	 *
+	 * @param IAppManager $appManager
+	 * @param IGroupManager $groupManager
+	 * @param IConfig $config
+	 */
+	public function __construct(IAppManager $appManager, IGroupManager $groupManager, IConfig $config) {
+		$this->appManager = $appManager;
+		$this->groupManager = $groupManager;
+		$this->config = $config;
+	}
+
+	/**
+	 *
+	 * @param IOutput $out
+	 *
+	 */
+	public function run(IOutput $out) {
+		//Get the whitelisted apps from the appconfig
+		$whitelistedApps = AppWhitelist::CORE_WHITELIST. ',' . $this->config->getAppValue(
+			'guests',
+			'whitelist',
+			AppWhitelist::DEFAULT_WHITELIST);
+
+		$appAccessService = $this->appManager->getAppsAccessService();
+
+		/**
+		 * Now copy the data from the appconfig to appaccess table.
+		 */
+		$groupObject = $this->groupManager->get(GroupBackend::DEFAULT_NAME);
+		$appAccessService->setWhitelistedAppsForGroup($groupObject, \explode(',', $whitelistedApps));
+
+		// Last step delete the data from the appconfig table
+		$this->config->deleteAppValue('guests', 'whitelist');
+	}
+}
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -47,7 +47,12 @@ class AppWhitelist {
 		if ($isGuest && $whitelistEnabled) {
 			$path = \OC::$server->getRequest()->getRawPathInfo();
 			$app = self::getRequestedApp($path);
-			$whitelist = self::getWhitelist();
+
+			//Get whitelisted apps
+			$appAccessService = \OC::$server->getAppManager()->getAppsAccessService();
+			$gid = \OC::$server->getConfig()->getAppValue('guests', 'group', GroupBackend::DEFAULT_NAME);
+			$group = \OC::$server->getGroupManager()->get($gid);
+			$whitelist = $appAccessService->getWhitelistedAppsForGroup($group);
 
 			if (!\in_array($app, $whitelist)) {
 				\header('HTTP/1.0 403 Forbidden');
@@ -58,17 +63,6 @@ class AppWhitelist {
 				exit;
 			}
 		}
-	}
-
-	public static function getWhitelist() {
-		$whitelist = self::CORE_WHITELIST;
-		$whitelist .=  ',' . \OC::$server->getConfig()->getAppValue(
-			'guests',
-			'whitelist',
-			self::DEFAULT_WHITELIST
-		);
-
-		return \explode(',', $whitelist);
 	}
 
 	/**
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -25,9 +25,12 @@
 namespace OCA\Guests\Controller;
 
 use OCA\Guests\AppWhitelist;
+use OCA\Guests\GroupBackend;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\IConfig;
+use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\IRequest;
 
@@ -54,11 +57,20 @@ class SettingsController extends Controller {
 	 */
 	private $l10n;
 
-	public function __construct($AppName, IRequest $request, $UserId, IConfig $config, IL10N $l10n) {
+	/** @var IAppManager */
+	private $appManager;
+
+	/** @var IGroupManager */
+	private $groupManager;
+
+	public function __construct($AppName, IRequest $request, $UserId, IConfig $config,
+								IL10N $l10n, IAppManager $appManager, IGroupManager $groupManager) {
 		parent::__construct($AppName, $request);
 		$this->userId = $UserId;
 		$this->config = $config;
 		$this->l10n = $l10n;
+		$this->appManager = $appManager;
+		$this->groupManager = $groupManager;
 	}
 
 	/**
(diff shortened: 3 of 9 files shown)


sharidas:
When https://github.com/owncloud/core/pull/36338 is merged, the acceptance tests should pass. I ran the webui tests. And it passed.

lib/AppWhitelist.php
```diff
@@ -61,13 +61,14 @@ public static function preSetup($params) {
...
-			'guests',
-			'whitelist',
-			self::DEFAULT_WHITELIST
-		);
+		$whitelist = self::CORE_WHITELIST . ',' . self::DEFAULT_WHITELIST;
 
+		$regulateAppAccess = \OC::$server->getAppManager()->getAppsAccessService();
+		$group = \OC::$server->getGroupManager()->get(GroupBackend::DEFAULT_NAME);
+		$result = $regulateAppAccess->getWhitelistedAppsForGroup($group);
+		if ($result === false || $result === []) {
```
jvillafanez: I see a couple of problems:
* If I want to use an empty whitelist for the guests, it will change to the default list.
* If I want to remove the whitelist and allow the guests to access to all the apps, it will again change to the default list.

I don't mind that the app to set a default list, but I'd want to keep control over the list.
Maybe a migration is easy enough to set the default whitelist, and then let the admin control it at will via the settings page.

lib/Controller/SettingsController.php
```diff
@@ -73,7 +86,9 @@ public function getConfig() {
 		} else {
 			$useWhitelist = false;
 		}
-		$whitelist = $this->config->getAppValue('guests', 'whitelist', AppWhitelist::DEFAULT_WHITELIST);
+		$appAccessService = $this->appManager->getAppsAccessService();
+		$whitelistedApps = $appAccessService->getWhitelistedAppsForGroup($this->groupManager->get(GroupBackend::DEFAULT_NAME));
+		$whitelist = \implode(',', $whitelistedApps);
```
jvillafanez: why the "implode + explode"?

lib/Controller/SettingsController.php
```diff
@@ -138,9 +153,33 @@ public function getWhitelist() {
...
+	 * Insert or update to the appacess table if the whitelisted apps are updated
+	 * This method would be called if the whitelisted apps are updated/reset to
+	 * default whitelist.
+	 *
+	 * @param array $apps
+	 * @return bool, true if data is inserted/updated to the appaccess table, else false.
+	 */
+	private function setOrUpdateWhitelist(array $apps) {
+		/**
+		 * Update or insert to db only if apps are not there for the group or
```
jvillafanez: Any problem just inserting directly? Performance-wise, you only need the "set" request, so it's faster than "get" + "set", and I don't see any real reason at the moment we want not to overwrite with the same list.

lib/Controller/SettingsController.php
```diff
@@ -24,10 +24,14 @@
 
 namespace OCA\Guests\Controller;
 
+use OCA\Files\App;
```
jvillafanez: This doesn't seem to be used

lib/Controller/SettingsController.php
```diff
@@ -138,9 +153,33 @@ public function getWhitelist() {
...
+	 *
+	 * @param array $apps
+	 * @return bool, true if data is inserted/updated to the appaccess table, else false.
+	 */
+	private function setOrUpdateWhitelist(array $apps) {
+		/**
+		 * Update or insert to db only if apps are not there for the group or
+		 * if apps are there but not same as DEFAULT_WHITELIST.
+		 */
+		$group = \OC::$server->getGroupManager()->get(GroupBackend::DEFAULT_NAME);
```
jvillafanez: This needs to be injected

lib/AppWhitelist.php
```diff
@@ -61,13 +61,14 @@ public static function preSetup($params) {
 	}
 
 	public static function getWhitelist() {
```
jvillafanez: This seems quite questionable... I'd use a `$appAccessService` and `$group` parameters in order to avoid having to rely on the `\OC::$server`. However, the method will likely just `$appAccessService->getWhitelist...($group)`, which is something the caller could do by itself.
Sound that this method should be removed. I barely see any advantage.

lib/Controller/SettingsController.php
```diff
@@ -54,11 +58,20 @@ class SettingsController extends Controller {
...
 	private $l10n;
 
-	public function __construct($AppName, IRequest $request, $UserId, IConfig $config, IL10N $l10n) {
+	/** @var IAppManager */
+	private $appManager;
+
+	/** @var IGroupManager */
+	private $groupManager;
+
+	public function __construct($AppName, IRequest $request, $UserId, IConfig $config,
```
jvillafanez: Check the formatting here.

lib/Controller/SettingsController.php
```diff
@@ -73,12 +85,12 @@ public function getConfig() {
 		} else {
 			$useWhitelist = false;
 		}
-		$whitelist = $this->config->getAppValue('guests', 'whitelist', AppWhitelist::DEFAULT_WHITELIST);
-		$whitelist = \explode(',', $whitelist);
+		$appAccessService = $this->appManager->getAppsAccessService();
+		$whitelistedApps = $appAccessService->getWhitelistedAppsForGroup($this->groupManager->get(GroupBackend::DEFAULT_NAME));
```
jvillafanez: question: any chance the group manager could return null?

lib/Controller/SettingsController.php
```diff
@@ -107,7 +119,7 @@ public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
 		$newWhitelist = \join(',', $newWhitelist);
 		$this->config->setAppValue('guests', 'group', $group);
```
jvillafanez: are these other "setAppValue" needed?

lib/Controller/SettingsController.php
```diff
@@ -138,9 +151,25 @@ public function getWhitelist() {
...
+	 * This method would be called if the whitelisted apps are updated/reset to
+	 * default whitelist.
+	 *
+	 * @param array $apps
+	 * @return void
+	 */
+	private function setWhitelistApps(array $apps) {
+		$group = $this->groupManager->get(GroupBackend::DEFAULT_NAME);
+		$appAccessService = $this->appManager->getAppsAccessService();
+		$appAccessService->setWhitelistedAppsForGroup($group, $apps);
```
jvillafanez: I think the service returns true if the data is saved in the DB. You can return that here and send a successful response if it's correctly saved and a wrong one if not.

appinfo/Migrations/Version20191115135305.php
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	public function run(IOutput $out) {
+		//Get the whitelisted apps from the appconfig
+		$config = \OC::$server->getConfig();
+		$whitelistedApps = AppWhitelist::CORE_WHITELIST. ',' . $config->getAppValue(
+			'guests',
+			'whitelist',
+			AppWhitelist::DEFAULT_WHITELIST);
+
+		$appManager = \OC::$server->getAppManager();
```
jvillafanez: Any chance to inject these services and include unittests for the migration?

lib/Controller/SettingsController.php
sharidas: It can if the backend doesn't exist.

lib/Controller/SettingsController.php
sharidas: Removed.

lib/Controller/SettingsController.php
sharidas: Done

lib/Controller/SettingsController.php
sharidas: Made change in the code, so that when group name is changed, we add to the backend so that `get` method called from groupManager doesn't fail.

lib/Controller/SettingsController.php
sharidas: Done.

appinfo/Migrations/Version20191115135305.php
sharidas: Injected the services. Also added the unit test for migration.

templates/settings/admin.php
```diff
@@ -30,9 +30,8 @@
 		<span class="inlineblock user-info-label"><?php p($l->t('Guest users are grouped under a virtual group in the user manager'));?></span><br/><br/>
 		<label for="guestGroup"><?php p($l->t('Group name'));?>
 		</label><input type="text" id="guestGroup" value="" /><br/>
-		<br/>
-		<input type="checkbox" id="guestUseWhitelist" value="useWhitelist"/><label for="guestUseWhitelist"><?php p($l->t('Limit guest access to an app whitelist'));?></label><br/>
-		<input type="text" id="guestWhitelist" value="" style="display:none; width:99%;"/><br/>
+		<label for="guestUseWhitelist"><?php p($l->t('Limit guest access to an app whitelist'));?></label><br/>
```
jvillafanez: this label doesn't seem right

templates/settings/admin.php
```diff
@@ -30,9 +30,8 @@
...
+		<input type="text" id="guestWhitelist" value="" style="width:99%;"/><br/>
```
jvillafanez: is the style really needed?

tests/unit/Version20191115135305Test.php
```diff
@@ -0,0 +1,82 @@
...
+		$appAccessService->expects($this->once())
+			->method('setWhitelistedAppsForGroup');
+		$this->appManager->method('getAppsAccessService')
+			->willReturn($appAccessService);
+
+		$groupObj = $this->createMock(IGroup::class);
+		$this->groupManager->method('get')
+			->willReturn($groupObj);
+
+		$this->invokePrivate($this->migrationVersionObj, 'run', [$ioutput]);
```
jvillafanez: The `invokePrivate` shouldn't be needed

lib/Controller/SettingsController.php
```diff
@@ -99,15 +113,34 @@ public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
...
-		$this->config->setAppValue('guests', 'group', $group);
-		$this->config->setAppValue('guests', 'usewhitelist', $useWhitelist);
-		$this->config->setAppValue('guests', 'whitelist', $newWhitelist);
+
+		if ($this->setWhitelistApps(\explode(',', $newWhitelist))) {
+			if ($group !== $oldGid) {
+				/**
+				 * Delete the old entry, else it becomes stale information.
+				 */
+				$this->appManager->getAppsAccessService()->wipeWhitelistedAppsForGroup($oldGid);
```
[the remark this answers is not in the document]
sharidas: Moved the code to the place where new group.

templates/settings/admin.php
sharidas: If we remove the style, then the text box shrinks and it would look ugly :(

tests/unit/Version20191115135305Test.php
sharidas: Ouch. Done.

templates/settings/admin.php
sharidas: Changed the label.

lib/Controller/SettingsController.php
```diff
@@ -99,15 +113,31 @@ public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
 				],
 			]);
 		}
+		$oldGid = $this->config->getAppValue('guests', 'group', GroupBackend::DEFAULT_NAME);
+		if ($group !== $oldGid) {
+			$this->config->setAppValue('guests', 'group', $group);
+			$newBackend = new GroupBackend($this->config, $group);
```
jvillafanez: The backend should be already registered at this point

lib/AppWhitelist.php
```diff
@@ -47,7 +47,11 @@ public static function preSetup($params) {
 		if ($isGuest && $whitelistEnabled) {
 			$path = \OC::$server->getRequest()->getRawPathInfo();
 			$app = self::getRequestedApp($path);
-			$whitelist = self::getWhitelist();
+
+			//Get whitelisted apps
+			$appAccessService = \OC::$server->getAppManager()->getAppsAccessService();
+			$group = \OC::$server->getGroupManager()->get(GroupBackend::DEFAULT_NAME);
```
jvillafanez: Note that the guest group name could have been changed

lib/Controller/SettingsController.php
```diff
@@ -99,15 +113,29 @@ public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
 				],
 			]);
 		}
+		$oldGid = $this->config->getAppValue('guests', 'group', GroupBackend::DEFAULT_NAME);
+		if ($group !== $oldGid) {
+			$this->config->setAppValue('guests', 'group', $group);
+			/**
+			 * Delete the old entry, else it becomes stale information.
+			 */
+			$this->appManager->getAppsAccessService()->wipeWhitelistedAppsForGroup($oldGid);
```
jvillafanez: we should also set the same whitelist for the new group

lib/Controller/SettingsController.php
```diff
@@ -124,10 +152,11 @@ public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
 	 * @return array whitelisted apps
 	 */
 	public function getWhitelist() {
+		$group = $this->groupManager->get(GroupBackend::DEFAULT_NAME);
```
jvillafanez: the group name could have been changed

lib/AppWhitelist.php
sharidas: Adjusted the change. The updated group is fetched.

lib/Controller/SettingsController.php
sharidas: Updated group is fetched now.

lib/Controller/SettingsController.php
sharidas: Moved the `$newWhitelist = [];` to the top of the method to avoid confusion. The same group is updated if the admin doesn't change the whitelist in the ui.

mmattel:
@sharidas any progress?

jvillafanez:
I think people decided for a different approach and this one died...

mmattel:
Totally ok, but then it should be closed and if possible referenced to the chosen one...

DeepDiver1975:
>  and if possible referenced to the chosen one...

https://github.com/owncloud/customgroups/pull/271

Issue #395, opened 2020-03-04
API acceptance tests do not work nicely for guest users with uppercase in the email


phil-davis:
The webUI test scenarios work fine for guest email addresses that contain uppercase, e.g. `John.Smith@example.com`

But the API test scenarios use code that tries to simulate what the webUI calls that are made to the guests app "private" API. The test code needs to be smarter about exactly when the upper or lower case version of the username/email-address needs to be put in various API calls and checked in responses.

e.g. in `apiGuests/guests.feature` replace `john.smith` with `John.Smith` - then sort out the test code so the tests pass.

kiranparajuli589:
> The webUI test scenarios work fine for guest email addresses that contain uppercase, e.g. `John.Smith@example.com`

@phil-davis IMO every guest emails are converted to lowercase during creation..
Our webUI test works fine because we also imitate naming convention using function `prepareUserNameAsFrontend` which lowers any upper-cased alphabets containing inside provided mail address and then continues to guest user creation

> But the API test scenarios use code that tries to simulate what the webUI calls that are made to the guests app "private" API. The test code needs to be smarter about exactly when the upper or lower case version of the username/email-address needs to be put in various API calls and checked in responses.

converting emails into lower-case may be bug or a feature,

kiranparajuli589:
Trying locally, when we create guest users with upper-cased email address:
![Screenshot from 2020-03-05 15-18-48](https://user-images.githubusercontent.com/39373750/75967983-c34f3280-5ef4-11ea-8c3e-e62a098c7e52.png)
- `username` is set to `lower-cased` mail address
- `email addr` is as we provided
- `fullName` email we provided

PR #398, opened 2020-03-05
[Tests-Only] API acceptance tests for guest users with uppercases in email

kiranparajuli589:
## Description
`username` of a `guest user` is set to `lower-cased email addr` during creation. So, `lower-cased` email should be used to share resource with guest user.

## Related Issue
- fixes https://github.com/owncloud/guests/issues/395

## Motivation and Context
share with guest users with emails containing upper-case letters

## How Has This Been Tested?
- :rabbit: 

## Types of changes

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiGuests/guests.feature
+++ b/tests/acceptance/features/apiGuests/guests.feature
@@ -14,7 +14,7 @@ Feature: Guests
     Examples:
       | email-address                  | user                 |
       | guest@example.com              | guest                |
-      | john.smith@email.com           | John.Smith           |
+      | John.Smith@email.com           | John.Smith           |
       | betty_anne+bob-burns@email.com | betty_anne+bob-burns |
 
   Scenario: Cannot create a guest if a user with the same email address exists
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/apiGuests/guests.feature
```diff
@@ -16,6 +16,7 @@ Feature: Guests
       | guest@example.com              | guest                |
       | john.smith@email.com           | john.smith           |
       | betty_anne+bob-burns@email.com | betty_anne+bob-burns |
+      | John.Smith@email.com           | John.Smith           |
```
phil-davis: Also try making the change in the Scenario Outline line 61
You can replace `john.smith` with `John.Smith` in lines 17 and 61 (no need to add a new examples line)

kiranparajuli589: @phil-davis Email addresses containing upper-cased letters: It's not a problem while creating guest user with such emails, but when we share resource to such user, then mail is sent to address that is lower-cased form of original email address.

PR #408, opened 2020-04-20
[Tests-Only] Update phpstan to latest version

jasson99:
Related - owncloud/QA#642

Update phpstan to 0.12

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -84,13 +84,12 @@ class SettingsController extends Controller {
 	/**
 	 * AJAX handler for setting the config
 	 *
-	 * @param $conditions string[]
-	 * @param $group string
-	 * @param $useWhitelist bool
-	 * @param $whitelist string[]
+	 * @param string $group
+	 * @param string $useWhitelist
+	 * @param string[] $whitelist
 	 * @return DataResponse
 	 */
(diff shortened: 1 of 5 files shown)


lib/Controller/SettingsController.php
```diff
@@ -84,13 +84,12 @@ public function getConfig() {
...
-	 * @param $group string
-	 * @param $useWhitelist bool
-	 * @param $whitelist string[]
+	 * @param string $group
+	 * @param string $useWhitelist
+	 * @param string[] $whitelist
 	 * @return DataResponse
 	 */
-	public function setConfig($conditions, $group, $useWhitelist, $whitelist) {
+	public function setConfig($group, $useWhitelist, $whitelist) {
```
phil-davis, line 92: @micbar @DeepDiver1975 @IljaN or whoever knows.
The first parameter `$conditions` is unused. `phpstan` reports that, and its easy to remove.
I see this method referenced in `appinfo/routes.php` but I don't see how it organizes how `setConfig` is called (i.e. with 4 or 3 parameters)

Is it OK to remove the unused parameter? Or do we need to keep it?

phil-davis:
@jasson99 please rebase and resolve conflicts. Then request review again and we can decide what to do next.

PR #489, opened 2022-03-14
add backend and javascript handling for domain block while sharing

mrow4a:
With this change we can exclude certain domains from guest sharing and these will not be displayed when sharing. 

**How to Configure?**
`occ config:app:set guests blockdomains --value 'test.com,gmail.com'`

**NOTE**
- this is currently configurable only via CMD, and admin implementation is missing
- this is only frontend implementation to disallow users from sharing with certain domains "by mistake", however this PR does not cover the blocking from backend and security aspect (thus one could circumvent JavaScript to create such guest share)


mrow4a:
I attempted to setup integration webui tests, but failed on doing so. I guess it would need to be done afterwards, but main point to test is something like this:

```
    Given user "Alice" has been created with default attributes and large skeleton files
    And the administrator adds config key "blockdomains" with value "oc.com" in app "guests" using the occ command
    And user "Alice" has logged in using the webUI
    And the user has opened the share dialog for file "lorem.txt"
    When the user types "Brian@oc.com" in the share-with-field
    Then user "Brian@oc.com" should not be displayed in dropdown as guest user
```

JanAckermann (approved):
LGTM

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIGuest-master-chrome-mysql8.0-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/guests/2350/12

mrow4a:
@phil-davis thanks, acceptance test look great!

js/guestshare.js
```diff
@@ -79,12 +79,43 @@
 		},
 	};
 
+	var GuestShareConfig = {
+
+		load: function () {
+			return new Promise(function (resolve, reject) {
```
jvillafanez, line 85:
```
return $.ajax({
  url: .....
});
```
The ajax call should fulfill the promise interface, so there is no need to wrap the call. No need for the success and error callbacks here because they'll be added later with the `then(...)`.
The `type` isn't needed (`GET` is the default), but better to use uppercase for consistency.

js/guestshare.js
```diff
@@ -79,12 +79,43 @@
...
 		OC.Plugins.register('OC.Share.ShareDialogView', {
 			attach: function (obj) {
 
+				// fetch guest share config, it is ok to get this async because
+				// if filtering wont be applied in frontend due to e.g. error it might
+				// be handled in the backend or entry not shown (depening on setting)
+				var guestShareConfig = null;
+				GuestShareConfig.load().then(
+					function (config) {
+						guestShareConfig = config;
```
jvillafanez, line 112: I'm not sure about this. If the `guestShareConfig` isn't needed in the line 166, why risking having a different behavior?, and if it's needed, I don't think this guarantees a "valid" `guestShareConfig`. We could get the config too late.

js/guestshare.js
mrow4a: @jvillafanez what you mention does not work, I would need to add setting "async: false" to ajax to make it work the way you explain. And I need this call async, would explain in other comment and also added more comments in a code regarding this. We use the same mechanism in wopi and I would prefer to keep it.

js/guestshare.js
```diff
@@ -132,6 +161,18 @@
...
+							if (guestShareConfig && guestShareConfig.blockdomains) {
+								for (i = 0 ; i < guestShareConfig.blockdomains.length; i++) {
+									if (searchTerm.endsWith('@' + guestShareConfig.blockdomains[i])) {
+										provideGuestEntry = false;
+									}
+								}
+							} else {
+								// if share config not ready or invalid config 
+								// do not provide entry
+								provideGuestEntry = false;
```
mrow4a, line 173: @jvillafanez added this here so that we do not add an entry to guest if config is still pending or invalid. this should not happen normally as call to get config is very fast, but on overloaded system could happen

js/guestshare.js
mrow4a: we add config async to avoid blocking sharing just because some stupid config did no load due to error etc.

mrow4a: of course, the small parts of the code that need this config should throw an error etc.

jvillafanez:
```
obj.autocompleteHandler = function(search, response) {
  return oldHandler.call(obj, search, function(result, xhrResult) {
    GuestShareConfig.promise().then(function (config) {
      .....
      if (config && config.blockdomains) {
        for (i = 0 ; i < config.blockdomains.length; i++) {
          if (searchTerm.endsWith('@' + config.blockdomains[i])) {
            provideGuestEntry = false;
          }
        }
      } else {
      .....
    };
  };
}
```
I was thinking something like that. That way, we ensure that the config is always loaded or we show an error message.

js/guestshare.js
jvillafanez: From https://api.jquery.com/jQuery.ajax/#jqXHR
> The jqXHR objects returned by $.ajax() as of jQuery 1.5 implement the Promise interface, giving them all the properties, methods, and behavior of a Promise (see [Deferred object](https://api.jquery.com/category/deferred-object/) for more information). These methods take one or more function arguments that are called when the $.ajax() request terminates.

I still don't see the need of wrapping the jqXHR object in a Promise when it's already a Promise (or acts as one)

mrow4a: I tested it, and it was weirdly behaving. I think we have this in other parts of the code and is battle-tested, so would prefer to keep it.

js/guestshare.js
mrow4a: @jvillafanez so generally, your suggestion would be to have the config loading being blocking for the guest sharing? meaning no config or invalid config = no guest sharing ?

jvillafanez: If you need the config, you should guarantee the config is loaded. If the config isn't loaded because the request fails, or the config has wrong data, you can handle that error the way you want.

The scenario I'd want to avoid is that you get the config after you run the autocomplete handler, so the autocomplete could wrongly assume that there is no config when in fact it's just the request didn't finish.
This might not happen with the current code because the handler might be executed way too late, so the config has plenty of time to finish, but in general I think it's a bad idea.

You might want to add some caching in the config request instead.
Under normal circumstances, the code could run as I've proposed, so when you run the autocomplete handler, you make a request for the config and once that request finish you go through the guest code.
If you want to preload the config, you can make a request for the config early, assuming the config will be cached transparently, so the next request (the one that would be performed in the autocomplete handler) will use the cached config instead of requesting a new one.
In this case, we're guaranteeing that if there is a configuration, the autocomplete handler will run with it.

js/guestshare.js
jvillafanez: Could you add additional information on the weird behavior and / or a link to explain the weird behavior?
Maybe we should some comment here explaining the situation. I mean, I can understand that you want to delay the resolution of a promise if you want to include additional processing of the config here, but if not I find it strange.

mrow4a:
@jvillafanez not to spend too much time on this async/failure-prone config loading, I decided to take "old school" way of loading that config that we use in all apps. Can you guys check again?

js/guestshare.js
```diff
@@ -132,6 +132,14 @@
 								}
 							}
 
+							if (oc_appconfig.guests && oc_appconfig.guests.blockdomains) {
+								for (i = 0 ; i < oc_appconfig.guests.blockdomains.length; i++) {
+									if (searchTerm.endsWith('@' + oc_appconfig.guests.blockdomains[i])) {
+										provideGuestEntry = false;
```
jvillafanez, line 138: we could break the loop earlier

jvillafanez (approved):
tiny improvement, but I don't think there will be a lot of domains anyway

PR #495, opened 2022-04-26
allow setting share block domains in admin panel

mrow4a:
fixes: https://github.com/owncloud/enterprise/issues/4990

Screenshot 2022-04-26 at 21 40 35

### DIFF
--- a/js/guests.js
+++ b/js/guests.js
@@ -30,6 +30,7 @@
 		var $section = $('#guests');
 		var $guestsByGroup = $section.find('#guestsByGroup');
 		var $guestGroup = $section.find('#guestGroup');
+		var $guestSharingBlockDomains = $section.find('#guestSharingBlockDomains');
 		var $guestUseWhitelist = $section.find('#guestUseWhitelist');
 		var $guestWhitelist = $section.find('#guestWhitelist');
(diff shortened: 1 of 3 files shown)


JanAckermann (approved):

phil-davis:
The string value is supposed to be a comma-separated list. But I can put in spaces between domain names, and I can enter any characters at all (that do not look like domain names)

We could:
1) do some easy automatic formatting (e.g. replace whitespace with commas, trim whitespace at beginning and end)
2) add some text to say that a comma-separated list of domain names is expected to be entered
3) (optional?) process the domain names through some "reasonable" validator to confirm that they are "valid" (the validator needs to accept less usual things like an IP address, "localhost" etc so that test/development installations can try local things)

mrow4a:
@phil-davis I am aware of points here, but in fact it was earlier agreed to do only CMD for this feature. However I noticed it was just easier to implement this than write docs. I think what you mention should be an issue in this repo. I will also add fixme in the code for this

pmaier1:
needs documentation @mmattel @EParzefall

pmaier1:
@mrow4a what will the user see if they enter a mail address with a domain that is on the block list?

Issue #498, opened 2022-05-02
Improve UX for guest sharing blocklist


pmaier1:
https://github.com/owncloud/guests/pull/495

Improve wording in the admin UI so that it is more self-explanatory. See proposal below and adapt it to the actual functionality, if necessary.

**Domain blocklist for guest user invitation**
The following domains will not be accepted when users try to invite guests. 
_Example: example1.com, example2.com, example3.com_

mrow4a:
@pmaier1 needs release

phil-davis:
> @pmaier1 needs release

See issue #499

PR #500, opened 2022-05-04
Added UI test scenario for blocked domains from sharing with guests

SagarGi:
## Description
This PR adds UI tests for sharing settings guests bloackdomains

## Related Issue

- Part of https://github.com/owncloud/guests/issues/496

## How Has This Been Tested?

- CI
- Locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #501, opened 2022-05-04
improve block domains admin setting

mrow4a:
Screenshot 2022-05-16 at 20 33 14

@pmaier1 as discussed in https://github.com/owncloud/guests/issues/498


pmaier1 (changes requested):
Thanks, but the sentence in between seems to be missing :)
> The following domains will not be accepted when users try to invite guests.

phil-davis:
@mrow4a PR #500 is merged and added a UI test for this block-domains UI element. The test only depends on the id of the element in the DOM, so it should still pass with the text changes here. Please rebase, and the test will run in CI of this PR.

mrow4a:
@pmaier1 updated and screenshot is in top comment

PR #502, opened 2022-05-17
for legacy hooks do not hint type

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/5193

we can hit exception as `{"reqId":"redacted","level":3,"time":"redacted","remoteAddr":"redacted","user":"--","app":"PHP","method":"GET","url":"\/core\/js\/oc.js?v=a56e4d21fce3a1c65e121c2dc61d270c","message":"Parameter 1 to OCA\\WOPI\\Hooks::extendJsConfig() expected to be a reference, value given at \/var\/www\/owncloud\/lib\/private\/legacy\/hook.php#103"}`

align with other apps like https://github.com/owncloud/core/blob/master/apps/files/lib/App.php#L46

### DIFF
--- a/lib/Hooks.php
+++ b/lib/Hooks.php
@@ -138,7 +138,14 @@ class Hooks {
 		}
 	}
 
-	public static function extendJsConfig(array &$array): void {
+	/**
+	 * Function used to extend global JS config emited with
+	 * OC_Hook::emit('\OCP\Config', 'js', ['array' => &$array]) and available
+	 * in JS as oc_appconfig.guests
+	 *
+	 * @param array $array holding $array['array'] key with a reference value to config
+	 */
+	public static function extendJsConfig($array): void {
 		$blockDomains = \OC::$server->getConfig()->getAppValue('guests', 'blockdomains');
 
 		$array['array']['oc_appconfig']['guests'] = [


lib/Hooks.php
```diff
@@ -138,7 +138,7 @@ public function handlePostShare(IShare $share) {
 		}
 	}
 
-	public static function extendJsConfig(array &$array): void {
+	public static function extendJsConfig($array): void {
```
jvillafanez, line 141: The `&` should be kept, otherwise the method won't do anything.

mrow4a: @jvillafanez you sure https://github.com/owncloud/core/blob/master/apps/files/lib/App.php#L46 ?

jvillafanez: If I have to spend 30 minutes checking it, at least I can say that the implementation is terrible.
* PHP arrays are passed by value, so my expectation here is that here you're modifying a copy of the array. The actual array shouldn't be modified. 
* The initial emit hook which ends up here has an "array" key with a reference as value (https://github.com/owncloud/core/blob/master/core/js/config.php#L269)
  * Assuming that this code works, I guess that modifying elements inside the "array" key will change the original data taking into account that's a reference.
  * The problem is that anything outside of that reference will stay in the copy. Adding a `$array['extra'] = ....` won't do anything. I'm pretty sure this will cause problems.
  * If the only usable key in the array is the "array" key, why would we need it? It's useless info that shouldn't be there.
* It would have been better to use an object instead of an array.

I'm not sure how much work is to fix this mess, but at least I think we should add a comment summing up the points above so the next time we check this code we know what's happening.

mrow4a: @jvillafanez yes, https://github.com/owncloud/core/blob/master/core/js/config.php#L269 is a core part. It is 6 years old code so I would not attempt to change it. But I agree, maybe comment would be good so that next person does not cause the same WARNING flood.

mrow4a: @jvillafanez adjusted description as agreed

jvillafanez (approved):

PR #506, opened 2022-06-21
Add batch action to allow sharing with multiple guests at once

JammingBen:
## Description
This feature allows sharing resources with multiple guests at once via the following format: *user1@mail.com;user2@mail.com;user3@mail.com*

Requires https://github.com/owncloud/core/pull/40155 to work. Merging this PR still keeps compatibility with older oC versions though.

## Related Issue
- Works towards https://github.com/owncloud/enterprise/issues/2865

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/js/guestshare.js
+++ b/js/guestshare.js
@@ -49,7 +49,7 @@
 			$.ajax(xhrObject).done(function (xhr) {
 				var properties = {
 					shareType: 0,
-					shareWith: self.email.toLowerCase(),
+					shareWith: email.toLowerCase(),
 					permissions: OC.PERMISSION_CREATE | OC.PERMISSION_UPDATE
 						| OC.PERMISSION_READ | OC.PERMISSION_DELETE
 				};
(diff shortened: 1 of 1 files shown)


mmattel:
Note that the requirement https://github.com/owncloud/core/pull/40155 has been merged

Any progress?

jvillafanez:
Should we use sets instead of arrays to hold the users? It should be more efficient than looping through an array multiple times, and the code might also be smaller and easier.

mrow4a:
I wonder, how error handling works here? If 1 user in a batch fails for whatever reason, what happens with a batch? how is this presented to the user?

JammingBen:
> Should we use sets instead of arrays to hold the users? It should be more efficient than looping through an array multiple times, and the code might also be smaller and easier.

@jvillafanez Do you mean here in L92: https://github.com/owncloud/guests/pull/506/files#diff-1e9748771effcb66e011a800b732cd0d897ea35001b2b40e9e9f7c970673ec7cR92?

> I wonder, how error handling works here? If 1 user in a batch fails for whatever reason, what happens with a batch? how is this presented to the user?

In an error case, all users who succeed will still be added. Failed users will be left out and prompted to the user:

image

jvillafanez:
Yes. I assume that both the result and the existing shares are arrays, and we can't change it, but we're creating an array for the users from the search. I'm not sure if we can skip the user loop somehow to take full advantage of the set operations. The goal would be to reduce the number of loops we need to do.

Anyway, the expected data set for the users, shares and result shouldn't be too big, so it could be ok.

JammingBen:
Okay got it, I already tried it in https://github.com/owncloud/customgroups/pull/521 (which is a similar care), unfortunately Sets do not support uniqueness with objects. Hence we would still need to loop through those. Or use some helpers from lodash, that probably do the same under the hood.

jvillafanez (approved):

jnweiger:
This is included in https://github.com/owncloud/guests/releases/tag/v0.11.0-rc.1

PR #527, opened 2022-10-07
Add encryption and files_primary_s3 to CORE_WHITELIST

phil-davis:
## Description
If encryption is used in an installation then it really needs to be always in the whitelisted apps for a guest, otherwise a guest is going to have trouble saving a file.

Similarly if files_primary_s3 is used in an installation then it really needs to be always in the whitelisted apps for a guest, otherwise a guest is going to have trouble saving a file.

So I think that these should be always part of the "core" whitelisted apps. There is not functionality that can be sensibly/usefully enabled and disabled for a guest user in these apps.

## How Has This Been Tested?
CI

## Types of changes

## Checklist:

### DIFF
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -30,7 +30,7 @@ use OCP\Template;
  * @package OCA\Guests
  */
 class AppWhitelist {
-	public const CORE_WHITELIST = ',core,encryption,files,files_primary_s3,guests';
+	public const CORE_WHITELIST = ',core,admin_audit,encryption,files,files_antivirus,files_primary_s3,firewall,guests,ransomware_protection';
 	public const DEFAULT_WHITELIST = 'settings,avatar,files_external,files_trashbin,files_versions,files_sharing,files_texteditor,activity,firstrunwizard,gallery,notifications,password_policy,oauth2,files_pdfviewer,files_mediaviewer,richdocuments,onlyoffice,wopi,oco_selfservice,twofactor_totp';
 
 	public static function preSetup($params) {
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -30,7 +30,7 @@ use OCP\Template;
  * @package OCA\Guests
  */
 class AppWhitelist {
-	public const CORE_WHITELIST = ',core,files,guests';
+	public const CORE_WHITELIST = ',core,encryption,files,files_primary_s3,guests';
 	public const DEFAULT_WHITELIST = 'settings,avatar,files_external,files_trashbin,files_versions,files_sharing,files_texteditor,activity,firstrunwizard,gallery,notifications,password_policy,oauth2,files_pdfviewer,files_mediaviewer,richdocuments,onlyoffice,wopi,oco_selfservice,twofactor_totp';
 
 	public static function preSetup($params) {
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -30,7 +30,7 @@ use OCP\Template;
  * @package OCA\Guests
  */
 class AppWhitelist {
-	public const CORE_WHITELIST = ',core,admin_audit,encryption,files,files_antivirus,files_primary_s3,firewall,guests,ransomware_protection';
+	public const CORE_WHITELIST = ',core,admin_audit,encryption,files,files_antivirus,files_external_s3,files_primary_s3,firewall,guests,ransomware_protection';
 	public const DEFAULT_WHITELIST = 'settings,avatar,files_external,files_trashbin,files_versions,files_sharing,files_texteditor,activity,firstrunwizard,gallery,notifications,password_policy,oauth2,files_pdfviewer,files_mediaviewer,richdocuments,onlyoffice,wopi,oco_selfservice,twofactor_totp';
 
 	public static function preSetup($params) {


phil-davis:
1) what other apps are like this, and should be added to the core whitelist?
Maybe all the storage-related apps??? (objectstore, files_external_*,...)
Maybe some of the "protection" apps - brute_force_protection, password_policy, firewall... - is it OK that an admin might have these set up and enable for the system, but they are not doing their "protection/enforcing" thing for guest accounts?

2) where should this "core whitelist" be stored? Is OK to be located here in the guests app itself? (that does make it easy to change and just quickly release the guests app)

phil-davis:
I added more apps to CORE_WHITELIST

jnweiger (approved):
That is definitly needed.
@pako81 what do you think?

jvillafanez (approved):

pako81:
Should not encryption only go into the CORE_WHITELIST while the rest would go into the DEFAULT_WHITELIST? We will probably want to keep the CORE_WHITELIST as simple as possible.

> 1. what other apps are like this, and should be added to the core whitelist?
Maybe all the storage-related apps??? (objectstore, files_external_*,...)
Maybe some of the "protection" apps - brute_force_protection, password_policy, firewall... - is it OK that an admin might have these set up and enable for the system, but they are not doing their "protection/enforcing" thing for guest accounts?

Yes, I would add the storage-related apps and it makes sense to add those protection apps as well as I don't see a use case why admin would not like to have those enabled for guest users (probably an admin would like guests to even stricter follow those protection policies).

My proposal:

```
CORE_WHITELIST = ',core,files,guests,encryption';

public const DEFAULT_WHITELIST = 'settings,avatar,files_external,files_trashbin,files_versions,files_sharing,files_texteditor,activity,firstrunwizard,gallery,notifications,password_policy,oauth2,files_pdfviewer,files_mediaviewer,richdocuments,onlyoffice,wopi,oco_selfservice,twofactor_totp,admin_audit,files_antivirus,files_primary_s3,firewall,ransomware_protection,windows_network_drive';
```

1 more app added to CORE_WHITELIST, 6 more to DEFAULT_WHITELIST. Objectstore is deprecated and replaced by files_primary_s3 so I would probably not add it. Maybe files_external, files_trashbin, files_versions, files_sharing should be moved to CORE_WHITELIST as they are actually part of core?

> 2. where should this "core whitelist" be stored? Is OK to be located here in the guests app itself? (that does make it easy to change and just quickly release the guests app)

There was some proposal to have a more sophisticated interaction between core and whitelisted apps for guest users. But for now I would probably leave it in the guests app itself as it make the release process easier.

phil-davis:
> What do you think?

If the used "storage" apps are not whitelisted for guests, then guest accounts just "do not work" - the underlying "storage driver" is not enabled for the guest, and so they cannot list or download the files that have been shared with them. So letting an admin not whitelist will just let the admin break their installation. IMO it is better to put such apps in the hard-coded whitelist.

pako81:
@phil-davis yes, I do agree, see my proposal above. Or did I miss some other storage apps? I would probably not add files_external_dropbox, files_external_ftp, files_ldap_home, files_onedrive, sharepoint, user_external by default as not used that often. files_external_s3 would probably makes sense, yes.

jvillafanez:
They need to be added in the core whitelist because we don't want the apps to be removed from the whitelist.
If userA shares a file in the ftp external storage with a guest, but the guest doesn't have the app whitelisted, I'm pretty sure that he won't be able to access to the file. It will look bad.

Apps in the default whitelist (not the core one) can be removed at any moment by the admin. We should make sure that, by removing any of those apps, we don't cause any problem. It doesn't matter if they're used often or not.

I agree this is not the best option and we should plan to improve this, but right now, this is the fastest solution we have.

pako81:
@jvillafanez good point, thanks. Then probably the storage-related ones should go in the core whitelist. Probably the security-related ones can stay in the default one as admin could change his/her mind and remove/add them at a later stage?

jvillafanez:
It depends on the type of app. I'd keep the same rule anyway: if the app mustn't be removed from the list, it has to be in the core whitelist.
I'm pretty sure admin_audit and firewall (among others) should also be included in the core whitelist. We don't want guests to bypass any security restriction.
I'm not sure how the password_policy app works, but maybe we can be a bit more lax with that one

phil-davis:
I created release issue #534

pako81:
New proposal:

```
public const CORE_WHITELIST = ',core,files,guests,encryption,files_primary_s3,files_antivirus,files_external,files_external_dropbox,files_external_ftp,files_ldap_home,files_onedrive,sharepoint,files_external_s3,admin_audit,firewall,ransomware_protection';

public const DEFAULT_WHITELIST = 'settings,avatar,files_trashbin,files_versions,files_sharing,files_texteditor,activity,firstrunwizard,gallery,notifications,password_policy,oauth2,files_pdfviewer,files_mediaviewer,richdocuments,onlyoffice,wopi,oco_selfservice,twofactor_totp,windows_network_drive';
```

@jvillafanez @phil-davis @jnweiger what do you think?

phil-davis:
I think that is adding these to the `CORE_WHITELIST`:
```
files_external,files_external_dropbox,files_external_ftp,files_ldap_home,files_onedrive,sharepoint,
```

That should not do any harm, and will help avoid potential problems for installations that use any of these "backend storage" apps.

@pako81 can you make a PR and we can review?

pako81:
https://github.com/owncloud/guests/pull/535

pako81:
https://github.com/owncloud/guests/pull/535 has been just merged. Closing this?

PR #600, opened 2023-09-06
Update AppWhitelist.php add always enabled apps to core whitelist

jnweiger:
Make guest's core_whitlist consistent with 
/core/shipped.json -> alwaysEnabled

### DIFF
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -30,7 +30,7 @@ use OCP\Template;
  * @package OCA\Guests
  */
 class AppWhitelist {
-	public const CORE_WHITELIST = ',core,files,guests,encryption,files_primary_s3,files_antivirus,files_external,files_external_dropbox,files_external_ftp,files_ldap_home,files_onedrive,sharepoint,files_external_s3,windows_network_drive,admin_audit,firewall,ransomware_protection';
+	public const CORE_WHITELIST = ',core,files,dav,federatedfilesharing,guests,encryption,files_primary_s3,files_antivirus,files_external,files_external_dropbox,files_external_ftp,files_ldap_home,files_onedrive,sharepoint,files_external_s3,windows_network_drive,admin_audit,firewall,ransomware_protection';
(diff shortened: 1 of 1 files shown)


phil-davis (approved):

cdamken (approved):
Tested and works!

jnweiger:
@phil-davis do you have an idea, why the whitelist starts with a comma? I cannot see a reason....

michaelstingl:
Should `oauth2` be `core` too?

jnweiger:
> Should `oauth2` be `core` too?

I don't think so.

phil-davis:
> @phil-davis do you have an idea, why the whitelist starts with a comma? I cannot see a reason....

The code in the guests app takes `CORE_WHITELIST` and uses the PHP `explode()` function to make an array of the comma-separated elements. That array is going to have a first element with key `0` and value the-empty-string. Looking at code that matches apps to the whitelist, it does not matter that this empty entry is in the array.

But it would be cleaner not to have the leading comma, and IMO everything will work fine without the leading comma.

jnweiger:
> > > Should `oauth2` be `core` too?
> > 
> > 
> > I don't think so.
> 
> So, when `oauth2` is enabled, server announces it, but then it's not working for guest users? Sounds like a bug to me. 🤔

I am talking about the hardcoded enable list. `oauth2` is part of default_whitelist, and should remain there. I'd leave that decision as a choice for admins. 

If you really want to hard code oauth2 as always enabled, the we would have to add it to the core whitelist, and have to remove it from the default whitelist to avoid confusion. But the second is only effective for first time installations. It has no effect (normally) when the value is taken from the DB, which s outside of our control. Thus it will remain listed in the sharing webpage, and leave the false impression, that it can be disabled.

michaelstingl (approved):

DeepDiver1975:
@jnweiger can you fix the typo in the commit message please?

DeepDiver1975:
oh - too late .....

I would really welcome a bit more care on commit messages ...... :cry:

PR #613, opened 2024-01-08
fix: respect default_language for guests invitation

pako81:
## Description
Respect default_language for guests invitation

## Related issue
https://github.com/owncloud/enterprise/issues/6322

## Motivation and Context
Currently, the `default_language` config.php option is wrongly not considered when generating templates for guests invitation. This has the consequence that the language of the user creating the guest is rather used, which may not be the same as the one used by the guest (especially true in case of guest users).

Note: the `sendGuestPlainInviteMail` method needed also to be adjusted as this is what is being used when creating a guest directly from custom groups --> https://github.com/owncloud/customgroups/blob/master/lib/Service/GuestIntegrationHelper.php#L124

## How Has This Been Tested?
Set `default_language` in config.php and observe this language is now used for guests templates.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -116,6 +116,13 @@ class Mail {
 			['fileId' => $share->getNode()->getId()]
 		);
 
+		$defaultLang = \OC::$server->getConfig()->getSystemValue('default_language', false);
+		// if set, use default_language in config.php for the guests template
+		if ($defaultLang !== false) {
+			$usedeflanguage = \OC::$server->getL10N('lib', $defaultLang);
+			$subject = (string)$usedeflanguage->t('%s shared »%s« with you', [$senderDisplayName, $filename]);
+		}
+
 		list($htmlBody, $textBody) = $this->createMailBody(
 			$filename,
 			$link,
@@ -123,7 +130,8 @@ class Mail {
 			$this->defaults->getName(),
 			$senderDisplayName,
 			$expiration,
-			$shareWithEmail
+			$shareWithEmail,
+			$usedeflanguage
 		);
 
 		try {
@@ -167,6 +175,13 @@ class Mail {
 
 		$subject = (string)$this->l10n->t('%s invited you', [$senderDisplayName]);
 
+		$defaultLang = \OC::$server->getConfig()->getSystemValue('default_language', false);
+		// if set, use default_language in config.php for the guests template
+		if ($defaultLang !== false) {
+			$usedeflanguage = \OC::$server->getL10N('lib', $defaultLang);
+			$subject = (string)$usedeflanguage->t('%s invited you', [$senderDisplayName]);
+		}
+
 		list($htmlBody, $textBody) = $this->createMailBody(
 			null,
 			null,
@@ -174,7 +189,8 @@ class Mail {
 			$this->defaults->getName(),
 			$senderDisplayName,
 			null,
-			$shareWithEmail
+			$shareWithEmail,
+			$usedeflanguage
 		);
 
 		try {
@@ -213,10 +229,11 @@ class Mail {
 	 * @param string $guestEmail
 	 * @return array an array of the html mail body and the plain text mail body
 	 */
-	private function createMailBody($filename, $link, $passwordLink, $cloudName, $displayName, $expiration, $guestEmail) {
+	private function createMailBody($filename, $link, $passwordLink, $cloudName, $displayName, $expiration, $guestEmail, $overrideL10n = null) {
 		$formattedDate = $expiration ? $this->l10n->l('date', $expiration) : null;
+		$l10n = $overrideL10n === null ? $this->l10n : $overrideL10n;
 
-		$html = new Template('guests', 'mail/invite');
+		$html = new Template('guests', 'mail/invite', '', false, $l10n->getLanguageCode());
 		$html->assign('link', $link);
 		$html->assign('password_link', $passwordLink);
 		$html->assign('cloud_name', $cloudName);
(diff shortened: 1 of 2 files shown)


lib/Mail.php
```diff
@@ -116,14 +116,23 @@ public function sendGuestInviteMail(Share\IShare $share, $uid, $token) {
 			['fileId' => $share->getNode()->getId()]
 		);
 
+		$usedeflanguage = null;
```
DeepDiver1975, line 119: Please use camel case naming
e.g. `$usedDefaultLanguage`

lib/Mail.php
```diff
@@ -167,14 +176,23 @@ public function sendGuestPlainInviteMail($shareWith, $uid, $token) {
 
 		$subject = (string)$this->l10n->t('%s invited you', [$senderDisplayName]);
 
+		$usedeflanguage = null;
```
DeepDiver1975, line 179: avoid duplication -> extract this to a method

DeepDiver1975: or better make it even part of the `createMailBody`function .....

DeepDiver1975 (changes requested):

lib/Mail.php
pako81: Not sure if this can be completely moved into the `createMailBody` function as we need to set the subject and this has to be done inside the `sendGuestInviteMail` and `sendGuestPlainInviteMail` methods. I have extracted the language check to a function to avoid duplication.

lib/Mail.php
```diff
@@ -116,14 +116,19 @@ public function sendGuestInviteMail(Share\IShare $share, $uid, $token) {
 			['fileId' => $share->getNode()->getId()]
 		);
 
+		if ($this->checkDefaultLanguage() !== null) {
```
DeepDiver1975, line 119: `check...()` implies that the function is checking some state or condition and returns true/false

but this function is actually returning some thing - `get...()` fits better

DeepDiver1975: in addition - let's save cpu cycles and call the method once - save the return in a variable and use it in the if condition and later in the if block as well as in the `createMailBody` argument.

lib/Mail.php
```diff
@@ -238,4 +249,17 @@ private function createMailBody($filename, $link, $passwordLink, $cloudName, $di
...
 		return [$htmlMail, $plainTextMail];
 	}
+
+	/**
+	 * check if default_language is defined in config.php
+	 * @return OCP\IL10N|null
+	*/
+	private function checkDefaultLanguage() {
+		$useDefaultLanguage = null;
+		$defaultLang = \OC::$server->getConfig()->getSystemValue('default_language', false);
```
DeepDiver1975, line 259: injection is missing ....

lib/Mail.php
DeepDiver1975: you can also remove the duplication of the subject generation

```php
$l10n = $this->getDefaultLanguage() ?? $this->l10n;
$subject = (string)$l10n->t('...')

**Review by DeepDiver1975 — CHANGES_REQUESTED** [review]

File: lib/Mail.php [hunk 5]
```diff
@@ -238,4 +251,16 @@ private function createMailBody($filename, $link, $passwordLink, $cloudName, $di
 
 		return [$htmlMail, $plainTextMail];
 	}
+
+	/**
+	 * get default_language is defined in config.php
```
pako81, line 256:
```suggestion
	 * get default_language if defined in config.php
```

lib/Mail.php
```diff
@@ -238,4 +251,16 @@ private function createMailBody($filename, $link, $passwordLink, $cloudName, $di
...
+	/**
+	 * get default_language if defined in config.php
+	 */
+	private function getDefaultLanguage() {
+		$useDefaultLanguage = null;
+		$defaultLang = $this->config->getSystemValue('default_language', false);
+		if ($defaultLang !== false) {
+			$useDefaultLanguage = \OC::$server->getL10N('lib', $defaultLang);
+		}
+		return $useDefaultLanguage;
```
phil-davis, line 264:
```suggestion
		return $defaultLanguage;
```
This is a string in the config like `en_GB` so the var name could be changed so that it does not sound like a boolean.
And add PHPdoc for the return type `?string`

lib/Mail.php
```diff
@@ -213,10 +225,11 @@ public function sendGuestPlainInviteMail($shareWith, $uid, $token) {
 	 * @param string $guestEmail
 	 * @return array an array of the html mail body and the plain text mail body
 	 */
-	private function createMailBody($filename, $link, $passwordLink, $cloudName, $displayName, $expiration, $guestEmail) {
+	private function createMailBody($filename, $link, $passwordLink, $cloudName, $displayName, $expiration, $guestEmail, $overrideL10n = null) {
```
phil-davis, line 228: Add PHP doc for the extra parameter.

lib/Mail.php
```diff
@@ -165,7 +174,9 @@ public function sendGuestPlainInviteMail($shareWith, $uid, $token) {
 
 		$this->logger->debug("sending invite to $shareWith: $registerLink", ['app' => 'guests']);
 
-		$subject = (string)$this->l10n->t('%s invited you', [$senderDisplayName]);
+		$useDefaultLanguage = $this->getDefaultLanguage();
```
phil-davis, line 177:
```suggestion
		$defaultLanguage = $this->getDefaultLanguage();
```
And adjust the var name in the code.

lib/Mail.php
```diff
@@ -101,7 +107,9 @@ public function sendGuestInviteMail(Share\IShare $share, $uid, $token) {
 		$this->logger->debug("sending invite to $shareWith: $registerLink", ['app' => 'guests']);
 
 		$filename = \trim($share->getTarget(), '/');
-		$subject = (string)$this->l10n->t('%s shared »%s« with you', [$senderDisplayName, $filename]);
+		$useDefaultLanguage = $this->getDefaultLanguage();
```
phil-davis, line 110:
```suggestion
		$defaultLanguage = $this->getDefaultLanguage();
```
And adjust the var name in the code.

pako81:
@phil-davis can you review this again? Thx!

DeepDiver1975:
in a perfect world we would cover this with unit tests ...

PR #617, opened 2024-01-22
fix: harden apis

DeepDiver1975:
## Types of changes

## Checklist:

### DIFF
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -49,7 +49,7 @@ class AppWhitelist {
 			$app = self::getRequestedApp($path);
 			$whitelist = self::getWhitelist();
 
-			if (!\in_array($app, $whitelist, true)) {
+			if (!\in_array($app, $whitelist)) {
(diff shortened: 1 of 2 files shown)


butonic (approved):

jvillafanez (approved):

DeepDiver1975:
Acceptance test 💥

DeepDiver1975:
@phil-davis mind having a look? THX

phil-davis:
> @phil-davis mind having a look? THX

I'm out for 2 hours, but will look again then.

`if (!\in_array($app, $whitelist, true))`

If I remove the `true` (strict) parameter from that, then the guest can login and tests pass.
At that point `$app` is "files" and "files" is in the `$whitelist` array, so the strict `in_array` should work just the same as without strict.
I need to look at what weird stuff is happening with https://www.php.net/manual/en/function.in-array.php in PHP 7.4

The test gets the guest logged in but the list of files is not shown:
Screenshot from 2024-01-23 13-58-08

PR #618, opened 2024-01-23
fix: detect files app in guest whitelist checking

phil-davis:
## Description
In `AppWhitelist.php` function preSetup:
```
			$path = \OC::$server->getRequest()->getPathInfo();
```
Path can be something like `/dav/files/guest@example.com/`

In that case, `getRequestedApp($path)` was returning `false`

Actually a path like that is to do with the files app. So return 'files'.

Then we can do `if (!\in_array($app, $whitelist, true))` (use a strict check that the app has to be in the whitelist array). This will help ensure that only requests to "app URLs" that are whitelisted will work for guests.

## How Has This Been Tested?
CI

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- [#618](https://github.com/owncloud/guests/pull/618) - fix: detect files app in guest whitelist checking
 
 ## [0.12.4] - 2024-01-23
 
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -49,7 +49,7 @@ class AppWhitelist {
 			$app = self::getRequestedApp($path);
 			$whitelist = self::getWhitelist();
 
-			if (!\in_array($app, $whitelist)) {
+			if (!\in_array($app, $whitelist, true)) {
 				\header('HTTP/1.0 403 Forbidden');
 				$l = \OC::$server->getL10NFactory()->get('guests');
 				Template::printErrorPage($l->t(
@@ -90,6 +90,8 @@ class AppWhitelist {
 			return 'heartbeat';
 		} elseif (\substr($url, 0, 13) === '/dav/comments') {
 			return 'comments';
+		} elseif (\substr($url, 0, 11) === '/dav/files/') {
+			return 'files';
 		}
 		return false;
 	}


phil-davis:
@DeepDiver1975 I found that `dav/files` was not being detected as a URL belonging to the "files" app. That was what prevented adding the strict check to `in_array`

DeepDiver1975:
nice catch - what strange logic .....

phil-davis:
And maybe it should be more generic to match any:
 `/dav/` is "dav" app.
`/webdav/` is also "dav" app.
?

Does it matter? Is there any use-case where the admin wants to whitelist "dav" but not "files", or whitelist "files" but not "dav"?
I doubt it - there is little point having guest users if they are not given permission to use both the "dav" and "files" apps.

phil-davis:
@DeepDiver1975 @jnweiger I rebased.
IMO this is "a good thing" and can be merged.

DeepDiver1975 (commented):
Reviewed the change — logic is correct.

1. `/dav/files/...` previously fell through `getRequestedApp()` to `false`; since `false` is never in the whitelist, guests hitting DAV files paths were 403'd. Mapping it to `'files'` fixes that — `files` is in `CORE_WHITELIST`, so it's always allowed. Good.
2. Adding the strict flag (`in_array($app, $whitelist, true)`) is a correct hardening: the whitelist is a string array, so an unmatched path returning `false` would previously loose-match (e.g. against `''`/`0`-like entries). Strict comparison means `false` matches nothing → unknown paths stay blocked. Both changes are sound.

CI:
- All GitHub Actions checks pass (PHPUnit ×4 DBs, PHP Code Style, JS Unit, semantic commits).
- The only red is `continuous-integration/drone/pr` (legacy Drone). Branch is BEHIND master — please rebase; the Drone failure looks unrelated to this 3-line change (GH Actions test suite is green), but confirm after rebase.

Nit: `CHANGELOG.md` has an `## [Unreleased]` section but no entry for this PR — please add one under `### Fixed` matching the existing `[#NNN](url) - <summary>` format.

phil-davis:
Rebased

phil-davis:
Changelog added.

DeepDiver1975 (commented):
LGTM. Re-push resolved the prior review points:

- CHANGELOG entry now present (`### Fixed` under `[Unreleased]`, links #618).
- Fix is correct: `getRequestedApp` maps `/dav/files/` → `files`, so the Files app is matched against the whitelist. Bonus: strict `in_array(..., true)` avoids type-juggling false positives.
- CI all green (the old red legacy check is gone).

Nothing blocking.

DeepDiver1975 (approved):

PR #633, opened 2024-08-14
test(refactor): fix nullable parameters error in nightly failure

anon-pradip:
## Description

This PR fixes nightly failures caused by `Implicit nullable parameters` issue

## Related Issue

Part of: https://github.com/owncloud/QA/issues/859

## How Has This Been Tested?

CI

## Types of changes

## Checklist:

### DIFF
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -69,7 +69,7 @@ class Mail {
 		IMailer $mailer,
 		Defaults $defaults,
 		IL10N $l10n,
-		IConfig $config = null,
+		?IConfig $config,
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
This is OK because `$config` is anyway not the last parameter, so removing the `= null` default loses nothing for us (the later parameters have to be specified anyway). For the current code and PHP 7.4 we don't need to think about calling the constructor with "named parameters" https://php.watch/versions/8.0/named-parameters

But in many cases we should also leave `= null` there so that a caller can leave out the parameter complete.ly, and it will default to `null`

PR #683, opened 2026-04-24
chore(ci): move basic webUI acceptance tests to GitHub workflows

phil-davis:
Note: there are other drone pipelines that run acceptance tests with scalityS3 and cephS3 storage in place.

Those need more setup to be able to run in a workflow, that will be done later.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -59,32 +59,6 @@ config = {
     "phan": False,
     "phpunit": False,
     "acceptance": {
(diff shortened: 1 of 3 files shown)


phil-davis:
Some of these acceptance tests need to send and read email.
ToDo: provide the option to run an email service in GitHub workflow CI.

saw-jan (approved):

Issue #91, opened 2017-05-24
Misleading actions in Users panel


cortho:
While the guests app creates a virtual group from the group appconfig value, in the admin frontend it looks like a normal group. So it is possible to make an admin out of a guest and vice versa even when the backend does not support this.

davitol:
![screen shot 2017-11-06 at 11 24 49](https://user-images.githubusercontent.com/10596560/32437004-36b29184-c2e6-11e7-82a2-a8c67552e71e.png)
Only webUI bug: It should not be possible to create a regular user and add it to a guest_group. (refreshing the browser we can see that the group is not set)

PR #74, opened 2017-04-25
Scripts in files app only to avoid collision with other vue apps

IljaN:

Issue #70, opened 2017-04-13
Improve invitation mailing


pmaier1:
The mailing that created guests receive should contain
- the respective instance URL in the subject 
(e.g. `ownCloud: ‘user’ shared ‘»file«’ with you on ‘InstanceURL’`)
- the info that the guest should use his mail address as username to log in

**Proposal**

`Hey there,`

`just letting you know that ‘user’ shared »file« with you on ‘InstanceURL’.`

`Activate your guest account at ownCloud by setting a password: http://localhost/index.php/lostpassword/reset/form/KgDStYk5ISaj3AO5OVtpC/bla`

`You can then log in using your e-mail address as username. Use this link to view the share: http://localhost/files/index.php?dir=%2F&scrollto=bla`

`Cheers!`
`--`
`ownCloud - Your Cloud, Your Data, Your Way!`
`https://owncloud.com`

Would also be nice not to have stuff like 'lostpassword'/'reset' in the activation URL

@IljaN

IljaN:
Should we really include a url to owncloud.com? (Branding?)

IljaN:
#57

IljaN:
ref: #57 #59

pmaier1:
> Should we really include a url to owncloud.com? (Branding?)

Well, I guess this is a template anyway which will be changed for branding, isn't it? @michaelstingl 
Otherwise you're right of course.

michaelstingl:
There are various parameters in the theme:

```
<?php p($theme->getName()); ?> -
<?php p($theme->getSlogan()); ?>
<br><a href="<?php p($theme->getBaseUrl()); ?>"><?php p($theme->getBaseUrl());?></a>
```
([core/templates/internalmail.php#L29-L31](https://github.com/owncloud/core/blob/master/core/templates/internalmail.php#L29-L31))

PR #692, opened 2026-07-15
fix(deps): resolve 3 qs Dependabot alerts (DoS)

DeepDiver1975:
## Description

Resolves all 3 open Dependabot alerts on `owncloud/guests`, all for the `qs` package (transitive, `development` scope).

`qs` is pulled in transitively via `body-parser@1.20.3 -> qs "6.13.0"` (karma's dev server dependency). It is pinned to an exact version by its only parent, so a plain upgrade cannot move it. Added a Yarn `resolutions` entry (`"qs": "^6.15.2"`); it now resolves to a single **6.15.3** copy, above all three advisory thresholds. `6.13.0 -> 6.15.3` is semver-compatible within the 6.x line, so `body-parser` keeps working.

### Fixed

| Alert | Package | Old → New | Advisory | Needs ≥ |
|---|---|---|---|---|
| #70 | qs | 6.13.0 → 6.15.3 | arrayLimit bypass in bracket notation → DoS via memory exhaustion (medium) | 6.14.1 |
| #74 | qs | 6.13.0 → 6.15.3 | arrayLimit bypass in comma parsing → DoS (low) | 6.14.2 |
| #91 | qs | 6.13.0 → 6.15.3 | `qs.stringify` crash on null/undefined in comma-format arrays with `encodeValuesOnly` → DoS (medium) | 6.15.2 |

### No fix available

None.

## Motivation and Context

Three remotely-triggerable DoS advisories against the bundled `qs`. Bumping past 6.15.2 clears all three.

## How Has This Been Tested?

- `yarn install` — lockfile re-resolves cleanly; `qs` collapses to a single `6.15.3` block.
- Verified `qs` loads at 6.15.3 and the previously-crashing comma-format / `encodeValuesOnly` path now handles `null` without throwing (the #91 regression).
- Verified `body-parser` resolves the single hoisted 6.15.3 copy — the resolution took effect and the only consumer still works.

Note: the full karma JS suite requires the guests app nested inside a *built* ownCloud core checkout (`core/vendor`), which isn't available in this environment. This change is lockfile-only; CI runs the full suite.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,9 @@
     "karma-junit-reporter": "*",
     "sinon": "^17.0.2"
   },
+  "resolutions": {
+    "qs": "^6.15.2"
+  },
   "engines": {
     "node": ">= 6.9",
     "yarn": ">= 1.0.0"
--- a/tests/acceptance/features/apiGuests/guests.feature
+++ b/tests/acceptance/features/apiGuests/guests.feature
@@ -495,7 +495,7 @@ Feature: Guests
     And as "Alice" file "/tmp/textfile0.txt" should not exist
     When user "Alice" tries to list the trashbin content for user "Alice"
     Then the HTTP status code should be "207"
-    And as "Alice" the file with original path "/tmp/textfile0.txt" should exist in the trashbin
+    And as "Alice" the file with original path "tmp/textfile0.txt" should exist in the trashbin
 
   @email @issue-553
   Scenario: guest user can delete shared files when files_trashbin app is whitelisted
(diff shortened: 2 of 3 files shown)


phil-davis (dismissed):

phil-davis:
There might be some change needed to explicitly whitelist the trashbin app.
I will have a look.

phil-davis:
Rebased to get the latest acceptance test workflow.

phil-davis (approved):

PR #697, opened 2026-07-21
fix: allow guests to complete login/2FA flow (AppWhitelist)

### LINKED ISSUE #696: Guest 2FA login broken: AppWhitelist blocks core /login/challenge/* route (regression from #618)
## Summary

Since #618 (`fix: detect files app in guest whitelist checking`, merged 2026-07-06), **guest users can no longer complete a two-factor-authentication login**. The core 2FA challenge route `/login/challenge/{provider}` is rejected with `403 Forbidden` by `AppWhitelist::preSetup`, so the verification page never loads.

This surfaced as a CI regression in `owncloud/twofactor_totp`: the `webUITOTPGuests` acceptance suite has been red on `master` and on every PR since then. cc @phil-davis (author of #618)

## Symptom

`tests/acceptance/features/webUITOTPGuests/guests.feature` — both scenarios fail deterministically at:

```
When the user re-logs in as "guest@example.com" to the two-factor authentication verification page
  Page\LoginPage::waitTillPageIsLoaded timeout waiting for page to load (Exception)
```

`owncloud.log` shows **no server exception** during the failing step — the page simply returns 403 and never renders, so Selenium times out.

Discriminating fact: the non-guest suite `webUITwoFactorTOTP` runs the *identical* re-login step successfully with a normal user. `AppWhitelist::preSetup` only runs when the session user `isGuest`, which is why the failure is guest-specific.

## Root cause

`lib/AppWhitelist.php`. #618 tightened the whitelist check from loose to strict comparison:

```php
-  if (!\in_array($app, $whitelist)) {          // loose (==)
+  if (!\in_array($app, $whitelist, true)) {    // strict (===)
```

1. During a guest's 2FA login, `preSetup` runs for the request to the core route `/login/challenge/{challengeProviderId}` (defined in `owncloud/core/core/routes.php` as `TwoFactorChallenge#showChallenge` / `#solveChallenge`).
2. `getRequestedApp('/login/challenge/totp')` has no branch matching `/login/*` (it handles `/apps/`, `/core/`, `/settings/`, `/avatar/`, `/heartbeat`, `/dav/comments`, `/dav/files/`), so it returns **`false`**.
3. `CORE_WHITELIST` begins with a leading comma (`',core,files,…'`), so `explode(',', …)` yields an empty string `''` as the first whitelist element.
4. **Before #618** — loose `in_array(false, [...])`: `false == ''` is `true` → the request was implicitly **allowed**.
5. **After #618** — strict `in_array(false, [...], true)`: `false === ''` is `false` → the request is **403 Forbidden**.

So the strict check (correct in intent) exposed a latent gap: `getRequestedApp()` doesn't recognise the core `/login/*` routes, and the leading-comma empty-string entry was silently papering over every unrecognised route under the old loose comparison.

## Suggested fix

Teach `getRequestedApp()` about the core login routes so guests can complete the login/2FA flow, e.g. add a branch mapping `/login` (and `/logout`) to `'core'` (already whitelisted):

```php
} elseif (\substr($url, 0, 6) === '/login' || \substr($url, 0, 7) === '/logout') {
    return 'core';
}
```

Two things worth deciding as part of the fix:
- **The leading-comma empty string** in `CORE_WHITELIST`/`DEFAULT_WHITELIST` should probably be dropped, or unrecognised routes (`getRequestedApp()` returning `false`) handled explicitly, so an empty string in the whitelist array can never accidentally allow/deny things.
- **A unit test** for `AppWhitelist` covering the login/2FA-challenge route for a guest would lock this in (there is currently no `AppWhitelistTest`).

## Impact

- Regression window: last green `twofactor_totp` master CI was run for commit #333 (2026-06-25); first red was #334 (2026-07-06, a CODEOWNERS-only change that cannot affect tests) — matching #618's merge time (2026-07-06 07:19Z).
- Any deployment with the guests app + 2FA enabled for guests is affected in production, not just CI.

## References
- Regression introduced by #618
- Core routes: `owncloud/core` → `core/routes.php` (`/login/challenge/{challengeProviderId}`)
- Failing CI suite: `owncloud/twofactor_totp` → `webUITOTPGuests`

DeepDiver1975:
## Description

Since #618 (`fix: detect files app in guest whitelist checking`) tightened
`AppWhitelist::preSetup` from a loose `in_array()` to a strict
`in_array(..., true)` check, guest users can no longer complete a
two-factor-authentication login: the core route `/login/challenge/{provider}`
is rejected with `403 Forbidden` and the verification page never renders.

Root cause: `getRequestedApp()` has no branch for the core `/login/*` routes,
so it returns `false`. Under the old loose comparison `false == ''` matched the
empty-string first element of `CORE_WHITELIST` (leading comma) and the request
was implicitly allowed; the strict comparison (`false === ''`) no longer
matches, so the request is denied.

This PR teaches `getRequestedApp()` about the core `/login` and `/logout`
routes, mapping them to the already-whitelisted `core` app so guests can
complete the login and 2FA flow. A new `AppWhitelistTest` locks in the
route→app mapping (there was previously no test for this class).

The latent empty-string entry in `CORE_WHITELIST` is intentionally left as-is:
under the strict comparison it is harmless, and every unrecognised route
correctly returns `false` and is denied.

## Related Issue

Fixes #696
Regression introduced by #618

## Motivation and Context

Any deployment with the guests app + 2FA enabled for guests is affected in
production, not just CI. This also fixes the red `webUITOTPGuests` acceptance
suite in `owncloud/twofactor_totp`, which has been failing on `master` and
every PR since #618 merged.

## How Has This Been Tested?

- Added `tests/unit/AppWhitelistTest.php` exercising the private
  `getRequestedApp()` via `TestCase::invokePrivate()` with a data provider
  covering all route mappings, including the `/login/challenge/totp` → `core`
  regression case and an unrecognised-route → `false` case.
- End-to-end: the `webUITOTPGuests` acceptance suite in `owncloud/twofactor_totp`
  is the real-world regression signal and should go green with this change.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- [#696](https://github.com/owncloud/guests/issues/696) - fix: allow guests to complete login/2FA flow (AppWhitelist)
 - [#618](https://github.com/owncloud/guests/pull/618) - fix: detect files app in guest whitelist checking
 
 ## [0.12.4] - 2024-01-23
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
   "name": "owncloud/guests",
   "config": {
     "platform": {
-      "php": "7.3"
+      "php": "8.3"
     },
     "allow-plugins": {
       "bamarni/composer-bin-plugin": true
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "79d45b32d25bb3f513bc1d4e1103863a",
+    "content-hash": "109af0351daa99d321224d95f892c175",
     "packages": [],
     "packages-dev": [
         {
@@ -73,7 +73,7 @@
     "platform": {},
     "platform-dev": {},
     "platform-overrides": {
-        "php": "7.3"
+        "php": "8.3"
     },
     "plugin-api-version": "2.9.0"
 }
--- a/lib/AppWhitelist.php
+++ b/lib/AppWhitelist.php
@@ -82,6 +82,8 @@ class AppWhitelist {
 			return  \OC_App::cleanAppId($app);
 		} elseif (\substr($url, 0, 6) === '/core/') {
 			return 'core';
+		} elseif (\str_starts_with($url, '/login') || \str_starts_with($url, '/logout')) {
+			return 'core';
 		} elseif (\substr($url, 0, 10) === '/settings/') {
 			return 'settings';
 		} elseif (\substr($url, 0, 8) === '/avatar/') {
--- a/lib/Controller/RegisterController.php
+++ b/lib/Controller/RegisterController.php
@@ -90,7 +90,7 @@ class RegisterController extends Controller {
 		IConfig $config,
 		ISecureRandom $secureRandom,
 		IMailer $mailer,
-		IUrlGenerator $urlGenerator
+		IURLGenerator $urlGenerator
 	) {
 		parent::__construct($appName, $request);
 
--- a/tests/unit/AppWhitelistTest.php
+++ b/tests/unit/AppWhitelistTest.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * ownCloud
+ *
+ * @copyright (C) 2026 ownCloud GmbH
+ * @license GPL-2.0
+ *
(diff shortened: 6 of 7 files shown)


phil-davis:
`fatal: unable to access 'https://codeberg.org/icewind/streams/': The requested URL returned error: 503`
https://github.com/owncloud/guests/actions/runs/29823191682/job/88610399884?pr=697

`codeberg` problems again!

https://codeberg.org/icewind/streams/
503 Service Unavailable
No server is available to handle this request.

The whole site is down.

phil-davis:
I added a commit to increase the PHP version in composer to 8.3, and PHPstan to major version 2. That makes it happy with the `str_starts_with` function.

phil-davis (approved):

